PR #11, opened 2018-02-03
Add first version of dutch translations

Rias500:
Did a preliminary translation of all fields, then I can see which translations are missing in the UI.


kjmartens:
Thank you! :)

PR #16, opened 2018-02-13
Fix #15 event ordering

### LINKED ISSUE #15: Adding new user throws  Undefined index: Freeform Pro

lenvanessen:
Fixed #15

### DIFF
--- a/src/Freeform.php
+++ b/src/Freeform.php
@@ -248,7 +248,10 @@ class Freeform extends Plugin
                         ],
                         self::PERMISSION_SETTINGS_ACCESS      => ['label' => self::t('Access Settings')],
                     ];
-                }
+                },
+                null,
+                false
+
             );
         }
 


gustavs-gutmanis:
@lenvanessen thanks for the PR.
I was wondering how could I test if this works as expected or not?  
I fixed the permissions problems by making the permission set use a static name for its key instead of a dynamic one, which was an oversight. And I get no more problems, and I was wondering if this PR is necessary once that's done.

lenvanessen:
Hi Gustavs,

My problem actually was related to 2 issues:
1. The $this->name is not the same as the identifier set by the non-pro version of the extension, thus resulting in an Identifier not found, this is fixed by the using the static var.
2. This issue still persists, I don't know why, maybe it's the order where the extensions are installed, but in my setup the event for the permissions is fired **before** the initial permissions are set by the non-pro version, thus it tries to merge an array that does not exists, resulting in the same issue. So by forcing the pro to append to the events, and the non-pro to not append, we make sure that the initial array is set first, before attempting to merge them into a new one.

gustavs-gutmanis:
Thanks for letting us know. I guess you might have installed pro before lite. No matter, I will make sure that whoever is initialized first is the one to set the permission block.
It will be fixed in the upcoming version.

lenvanessen:
I think you can close this as well: https://github.com/solspace/craft3-freeform-pro/pull/1

gustavs-gutmanis:
Yeah, thanks again!
I'm closing this PR since we'll use a bit different approach here.

PR #40, opened 2018-04-04
Fixed a bug where the Craft edition referenced no longer existed.

andris-sevcenko:


andris-sevcenko:
Seems to be moot now that the fixes are in. Also wrong now that I take another look. Should have been `>= \Craft::Pro`

PR #55, opened 2019-04-04
Add getTagAttributes() function to Form component

peterhicks:
This PR implements a getTagAttributes() onto the Form component.

Right now, you can't access the tag attributes on a form when accessing it programatically, such as thorough an event. As it is one of the simplest ways to store arbitrary data that is editable in the admin panel and theoretically accessible by both the front-end and back-end, it would make sense to make it accessible through the Form object like other attributes.


kjmartens:
Thanks! We'll have a developer review this soon. :)

kjmartens:
Will be available in Freeform v2.5.17 shortly

PR #57, opened 2019-08-02
Added german translations

sarahschuetz:


kjmartens:
@sarahschuetz Thanks for this! We'll add this to the next version of Freeform. :)

PR #58, opened 2019-08-14
Honeypot html validation

DiederikVanHoorebeke:
`aria-hidden="true"` was set twice


kjmartens:
@DiederikVanHoorebeke Thank you... this will be included in the next release. :)

PR #59, opened 2019-12-10
Allow reCAPTCHA keys to be set via environment variables

NRauh:
If a project is using the project config, the value of `recaptchaKey` and `recaptchaSecret` could be committed when it shouldn't.

These changes make it so the value could be something like `$RECAPTCHA_SECRET_KEY` and the value of the environment variable `RECAPTCHA_SECRET_KEY` would be used.

From testing this out I noticed that the ReCAPTCHA field needs to be readded to the form before using an environment variable, otherwise it will still use the literal value.

I don't believe this would cause any problem for users updating the plugin, since this is the current behavior. But it's something worth mentioning to users who want to switch.

Let me know if I missed any usages of `recaptchaKey` or `recaptchaSecret`.

Here are a couple links with more info

- https://docs.craftcms.com/v3/config/environments.html
- https://docs.craftcms.com/v3/extend/environmental-settings.html

### DIFF
--- a/src/Fields/RecaptchaField.php
+++ b/src/Fields/RecaptchaField.php
@@ -39,7 +39,7 @@ class RecaptchaField extends AbstractField implements NoStorageInterface, Single
         /** @var Settings $settings */
         $settings = Freeform::getInstance()->getSettings();
 
-        $key   = $settings->recaptchaKey;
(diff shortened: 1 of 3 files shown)


kjmartens:
Added this into v3.5.9, thank you!

PR #60, opened 2020-02-16
Fix Quick Export error

dreamseeker:
'freeform_export_settings' table will be empty after restoring with './craft restore/db' command.

So next time I open the 'Quick Export' modal, checkboxes in the field list do not appear.

This commit is a modified code to solve it.

- Craft CMS 3.4.5
- Freeform Pro 3.6.2


kjmartens:
Thanks @dreamseeker. We've added a similar fix into 3.6.4. :)

PR #61, opened 2020-03-02
Improve file upload security and information disclousure on error

darksaboteur:
The PR checks the files mime type (if supported by the server) and converts it to a file extension. If the mime type is not in the allowed extension list an error will be thrown.

For example this stops .txt files being renamed to .jpg to pass the simple extension check.

The other change in src/Services/FilesService.php returns a generic error to the user rather than potentially sensitive server information, in our case it was returning details AWS or local file filesystem details.


kjmartens:
Included for next release (3.6.6), thank you very much!

PR #63, opened 2020-03-20
array_key_exists is deprecated in PHP 7.4

incraigulous:
Note that you should also merge in https://github.com/solspace/craft3-freeform/pull/62 to completely solve the issue.


kjmartens:
Thanks! We'll tackle this for Freeform 3.6.7 shortly.

kjmartens:
Added in v3.6.7 :)

PR #65, opened 2020-03-27
Add hook for other plugins

urbantrout:
Give plugins a chance to add other things here. Just like Craft CMS does: https://github.com/craftcms/cms/blob/62b9b99fc78b69a127db2d2c9003fd4072f15d3b/src/templates/entries/_edit.html#L147

### DIFF
--- a/src/templates/submissions/_edit_base.html
+++ b/src/templates/submissions/_edit_base.html
@@ -541,6 +541,8 @@
             </div>
         </div>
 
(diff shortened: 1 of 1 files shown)


kjmartens:
Thanks! We'll review and consider this for Freeform 3.6.7 shortly.

kjmartens:
Added in v3.6.7 :)

PR #67, opened 2020-08-12
Allow Symfony Filesystem ^5.0

ccchapman:
This version is now indirectly required by Craft 3.5, due to a `composer/composer` requirement.

No found usages of either changed methods in `symfony/filesystem`.

**Refs:**

- [symfony/filesystem/blob/master/CHANGELOG.md#500](https://github.com/symfony/filesystem/blob/master/CHANGELOG.md#500)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     "craftcms/cms": "^3.4.0",
     "symfony/property-access": "^2.8|^3.0|^4.0",
     "symfony/finder": "^2.8|^3.0|^4.0",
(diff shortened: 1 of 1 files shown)


kjmartens:
Thank you for your time! We're already about to release an update for this :)

ccchapman:
Thank you @kjmartens. It seems like some of the other Symfony packages also need version constraint changes to be compatible, but I'll leave it to you!

kjmartens:
@ccchapman Thanks! We'll get that fixed too

PR #68, opened 2020-08-13
Check empty getAllowedSubmissionFormIds() return

jeroenverfallie:
`getAllowedSubmissionFormIds()` can be null in case the user has access to all forms.
https://github.com/solspace/craft3-freeform/blob/e52a1da0be90ee057387c9abd5acb2ad21615b26/src/Services/SubmissionsService.php#L438-L452

This causes an error here:
https://github.com/solspace/craft3-freeform/blob/e52a1da0be90ee057387c9abd5acb2ad21615b26/src/Elements/Submission.php#L224

Error reproduction can be done using these permissions:
image


kjmartens:
Thanks! We just discovered this issue last night and have a fix coming out for it today. :)

PR #70, opened 2020-09-16
Add country and job title field support to Pardot integration

tomdavies:
While integrating Freeform Pro on a client site and making use of your Pardot integration (thanks for that BTW), we had need of some of the additional fields that Pardot supports out of the box but which the Freeform integration does not support mapping form fields to.

This PR adds the ability to map the `job_title` and `country` fields, as that's what we needed for our use case, however if it would help to get the PR merged I'd be happy to add support for the [full set of additional fields Pardot supports setting on a Prospect when creating](https://developer.pardot.com/kb/object-field-references/#prospect) as it's not a lot of work to do.


kjmartens:
This has been added in 3.9 now :)

PR #72, opened 2020-12-10
Add multiselectpicklist as a Zoho Lead field type

mdominguez:
Add multiselectpicklist as a Zoho Lead field type + configure to be sent as an array instead of falling back to be sent as a string


gustavs-gutmanis:
Thanks @mdominguez

PR #104, opened 2021-02-13
Add support for Twig in admin recipients field

engram-design:
A client of ours requires the use of Twig in the Admin Recipients field, so allow them to setup custom (complex) logic to add recipients based on a number of different field values. This PR adds Twig-parsing of submission variables.

This was achievable through the `EVENT_BEFORE_SEND` event in the Craft 2 version of Freeform, but this Craft 3 version doesn't support the ability to modify the recipients, as they're already processed before reaching the event trigger, and can't be modified.

### DIFF
--- a/packages/plugin/src/Library/Composer/Components/Properties/AdminNotificationProperties.php
+++ b/packages/plugin/src/Library/Composer/Components/Properties/AdminNotificationProperties.php
@@ -39,7 +39,7 @@ class AdminNotificationProperties extends AbstractProperties
     /**
      * Gets all recipients as an array.
      */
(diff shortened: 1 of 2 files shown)


gustavs-gutmanis (changes requested):
Thanks @engram-design, a very nice addition.
Please address the two small changes I mentioned and I'll merge this PR.

engram-design:
No worries, happy to tweak this. My only concern was with Will bull able types and the PHP compatibility. I know it can be a hit finicky!

gustavs-gutmanis:
Thanks

PR #198, opened 2021-08-26
updated Sharpspring crm integration, custom fields did not show up in cms

ArneVanBael:
Recently tested the SharpSpring CRM integration in the cms of the website of our company and noticed the SharpSpring custom fields didn't show up.
It appears that the sharpspring api has changed, the method 'getFields' returns all the fields including the standard fields.
Show we can add the 'isCustom = 1' filter in the api request.
Next, the properties 'hidden', 'readonlyvalue', 'calculated' do not exist anymore.
Also the $field object wasn't casted to an object which triggered the first condition and resulted in an empty list instead of all the custom fields.

### DIFF
--- a/packages/plugin/src/Integrations/CRM/SharpSpring.php
+++ b/packages/plugin/src/Integrations/CRM/SharpSpring.php
@@ -133,7 +133,7 @@ class SharpSpring extends AbstractCRMIntegration
      */
     public function fetchFields(): array
     {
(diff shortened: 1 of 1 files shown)


kjmartens:
Thanks for this @ArneVanBael. We'll look it over and add it in soon. :)

gustavs-gutmanis:
Thanks for the PR @ArneVanBael!

PR #350, opened 2022-05-17
Fixed a SQL error that would occur on PostgreSQL when viewing form submissions.

angrybrad:


kjmartens:
Thank you for this @angrybrad. We'll review and add this in shortly. 🙂

PR #404, opened 2022-07-22
Use new Campaign integration method

bencroker:
I've made it easier to integrate with the Campaign plugin going forward. This PR uses the new `FormsService::createAndSubscribeContact` method, falling back to the manual way of doing things if the method does not exist (if an older version of Campaign is installed). I also added a `TODO` to remind you to remove the manual approach in the next major version of Freeform.

### DIFF
--- a/packages/plugin/src/Integrations/MailingLists/Campaign.php
+++ b/packages/plugin/src/Integrations/MailingLists/Campaign.php
@@ -96,40 +96,46 @@ class Campaign extends AbstractMailingListIntegration
         }
 
         foreach ($emails as $email) {
-            $contact = CampaignPlugin::$plugin->contacts->getContactByEmail($email);
-
-            if (null === $contact) {
-                $contact = new ContactElement();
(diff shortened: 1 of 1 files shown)


kjmartens:
Thanks @bencroker. We'll check into it. 🙂 

Just to confirm, this only applies to Campaign 2 for Craft 4, therefore only Freeform 4 for Craft 4?

bencroker:
That's correct, Kelsey. Hopefully I PR'd into the correct branch?

kjmartens:
You PR'd the FF3 branch, but no worries... we'll sort it out. Thanks for the for assistance. 🙂

bencroker:
Ah, my mistake, I'll correct it, but what is the best branch to use, `5.0`?

kjmartens:
v4.0/master 🙂

bencroker:
Closing in favour of https://github.com/solspace/craft-freeform/pull/412

PR #412, opened 2022-07-27
Use new Campaign integration method

bencroker:
I've made it easier to integrate with the Campaign plugin from v2.1.0. This PR uses the new `FormsService::createAndSubscribeContact` method, falling back to the manual way of doing things if the method does not exist (if an older version of Campaign is installed). I also added a `TODO` to remind you to remove the manual approach in the next major version of Freeform.


seandelaney:
Tested and all worked as expected.

Issue #437, opened 2022-08-23
Active Campaign email marketing integration tags not posting correctly


DieterVercammen:
**Describe the bug or issue you're experiencing**
Active Campaign email marketing integration does not post existing tags correctly to AC.

**Steps to reproduce**

1. Setup an AC email marketing integration (with more than 20 existing tags)
2. Add a hidden field to your form to use as a mapper for the tags you want to post.
3. Set the default value of your hidden field to multiple tags (e.g. Tag#1;Tag#2;Tag#3)
4. Map the hidden field to the AC tags though the AC email marketing integration
5. Submit form

**Expected behavior**
All tags should be pushed to AC.

**Craft & Plugin Info (please complete the following information):**

- Craft Version: 3.7.51
- Freeform Version: 3.13.14
- Freeform Edition: PRO
- Fresh Install or Upgrade: Fresh install

**Additional context**
Once Active campaign has more than 20 tags, you will have to paginate the request that fetches existing tags. Otherwise the code will push a new tag to AC which already exists.

kjmartens:
Sorry for the trouble, and thanks for the PR @DieterVercammen. We'll review and add it soon. 🙂

geertvansoest:
@kjmartens This issue seems to be fixed in [3.13.19](https://github.com/solspace/craft-freeform/releases/tag/v3.13.19), but not for Craft 4 (we are currently using Freeform 4.0.7). Is it possible to enroll the same fix for 4.x?

https://github.com/solspace/craft-freeform/compare/v4.0.7...v3.13.19#diff-ba6921acca410a748d21abff66f6b710cb07da8ff99494f0d454285f853fa78c

kjmartens:
Sorry @geertvansoest. We somehow missed this one for v4. We'll add it shortly. 🙂

PR #438, opened 2022-08-23
fix: add private method to fetch existing tags with pagination

DieterVercammen:
Fix for https://github.com/solspace/craft-freeform/issues/437


DieterVercammen:
Hi @gustavs-gutmanis 

Thanks for the quick response and you're very welkom for the PR. 
I've tested the changes and updated 1 typo and updated the initial offset for the existing tags. Looks like everything is working correctly

Issue #485, opened 2022-10-26
Deleting form submissions crashes and runs out of memory


darylknight:
**Describe the bug or issue you're experiencing**
We're using Freeform as an age gate for an alcohol website. I need to "save submissions to the database" (form setting), because I'm using the most recent submission on the template to check the submitted date is more than 21 years ago. 

Two separate issues / requests:

1. I don't actually want those submissions in the database, but I can't check the submitted date if this isn't enabled. It would be great if we could set Freeform to automatically purge submissions to that form after X days, just like the spam system works.
2. We have 33,000 submissions to that form. If I go to Submissions and select Delete All, it give sme a server error because it immediately runs out of memory. Perhaps this could be avoided if Freeform tried to batch that operation into smaller chunks instead of requesting to do all 33k in one go?
 
![image](https://user-images.githubusercontent.com/1250017/198012006-878930aa-e01b-45f7-afe9-328370bedb80.png)

```
[26-Oct-2022 04:10:18 America/Los_Angeles] PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 8192 bytes) in /home/ploi/sitename.com/vendor/yiisoft
```

**Steps to reproduce**

1. Have lots of form submissions
2. Try to delete them all in one go on a small server

**Expected behavior**
Freeform should batch this operation so it doesn't run out of memory

**Craft & Plugin Info (please complete the following information):**

- Craft Version: 4.3.0
- Freeform Version: 4.0.7
- Freeform Edition: Pro

darylknight:
If you're suggesting writing a PHP module, no that wouldn't be possible for us. We're currently doing it like this:

```
{# * Make sure AJAX is disabled on the age gate form #}
{# * Make sure the form is set to "save submissions" - otherwise the below code breaks (it relies on checking the previous submission) #}

{% set form = craft.freeform.form('ageGate') %}

{% if getCookie('ageGateCookie') %}
  {# ? There is a cookie set, which means they've passed the age gate within the last 30 days #}
{% else %}
  {# No cookie set, so check the age #}
  {% if form %}
    {# Check if the form exists before doing any of this #}
    {% if form.submittedSuccessfully %}
      {# form was submitted successfully #}
      {% set submission =
        craft.freeform.submissions({
          form: 'ageGate'
        }).one()
      %}
      {% if submission %}
        {% if (submission.dateOfBirth|date('U')) > ('-21 years'|date('U')) %}
[...]

PR #488, opened 2022-10-31
[SFT-9]: Deleting thousands of submissions runs out of memory

gustavs-gutmanis:
Adresses issue #485

* fixed PHP running out of memory when deleting thousands of submissions at once by refactoring the delete process to use batching


seandelaney (approved):
Code changes looking good. Really like the move over to query batching.

PR #490, opened 2022-11-02
[SFT-29]: Persist: form data

gustavs-gutmanis:
# Description

* saving forms calls either `POST /api/forms` or `PUT /api/forms/:id` endpoints
* it's either successful on submit (`201` or `204`) or throws an error with error messages (`400`)
* errors are persisted in the form state


packages/client/config/axios/APIError.ts
```diff
@@ -0,0 +1,19 @@
...
+export type ErrorList = {
+  [key: string]: string[];
+};
+
+type ErrorCollection = {
+  [key: string]: ErrorList;
+};
+
+export class APIError extends Error {
+  errors: ErrorCollection = {};
```
seandelaney, line 12: `name: string = null;` ?

seandelaney: We have errors defined (`errors: ErrorCollection = {};`) but not `name`.

seandelaney (approved):
Looks fine.

I think all the PHP changes are mainly cleanup/formatting/adding types etc.

TSX/TS changes are looking great upon visual review. Still need to wrap my head around how this works before I can start suggesting improvements or changes (if there was any).  It does show me that I need to update my Settings work to follow suit - e.g the persisting data part.

PR #533, opened 2022-12-02
[SFT-89]: reCAPTCHA scripts load on ANY page that includes a Freeform form, regardless of whether the form(s) have reCAPTCHA enabled

seandelaney:
- Fixed undefined error (using `PaymentInterface` instead of `PaymentFields`)
- Fixed bug where `disableRecaptcha` form parameter was only considered in the Checkbox setup and not Invisible setup.
- Updated the JS to prevent the ReCaptcha API script loading multiple times when multiple forms on the same page.
- Flipped some logic used to process ReCaptcha after form submission.
- Cleaned up some logic too.


seandelaney:
@gustavs-gutmanis @kjmartens PR updated again. Had to flip the logic to get the `disableRecaptcha: true,` form property to work in all cases.

PR #544, opened 2022-12-09
[SFT-90]: Support sameSiteCookieValue

seandelaney:
Fixed error caused by trailing comma when running PHP 7.2.


kjmartens (approved):
Tested and works :)

PR #549, opened 2022-12-13
[SFT-57]: Refactor the `AbstractProperty` instances into `#[Property]`

gustavs-gutmanis:
Refactored all property handling in forms

Now, instead of defining each property in the `Form` object, and having to create a `SomeProperty` class which extends `AbstractProperty` class that defines all of the properties in it and exposes getters and asks for `manifest` that defines the types of all property values - you create a specific `SettingsNamespace` class which has all of the properties defined in it, and every one of them is described with a `#[Property]` attribute.

That instead generates all of the tabs and properties and sections in the form builder and upon saving, it validates everything and puts in in their respective places.


packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/field-layout/cell/cell-types/cell-field/cell-field.tsx
```diff
@@ -28,7 +27,9 @@ export const CellField: React.FC<Props> = ({ uid }) => {
       }}
     >
       <Label>{field.properties.label || type?.name}</Label>
-      <Text />
+      <div>
+        <input type="text" className="text fullwidth" />
+      </div>
```
gustavs-gutmanis, line 32: This was and still is a temporary placeholder in the builder layout view. Do not mind this.

seandelaney (approved):
Fantastic changes in here!

PR #552, opened 2022-12-15
[SFT-27]: Implement property value editing middleware functionality

gustavs-gutmanis:
* added `applyMiddleware` functionality which lets us add additional value transformations without touching react eventually


seandelaney (approved):
Magic!! Harry Plotter should be getting worried.

PR #553, opened 2022-12-15
[SFT-97]: Property type: option (Manual Entry of Options)

seandelaney:
- Added new `Options` component
- Within the `Options` component exists a new `OptionsEditor` component, that manages the different types of `sources`.
  - The `OptionsEditor` allows users to select a `source` (only Custom Options are supported for now).
  - The `CustomOptions` component provides a neat little UI to build label/value lists that can be then used within the `Select `component.

- Neat features include:
  - Setting auto focus on last available label input field when adding new option.
  - Pressing enter key within any text input fields adds a new option.


packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.styles.ts
```diff
@@ -0,0 +1,286 @@
+import styled from 'styled-components';
+
+export const Wrapper = styled.div`
+  padding: 0;
+  width: 100%;
```
gustavs-gutmanis: Please only write rules that are necessary.
I'm not sure about the `padding`, maybe it's required here, but `width` is always 100% for div elements, so that's a redundant rule.
And I'm not sure that `height: 100%` is also necessary here.

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.styles.ts
```diff
@@ -0,0 +1,286 @@
...
+  height: 100%;
+  display: flex;
+  font-weight: bold;
+  margin: 0 0 10px 0;
+  flex-direction: column;
+  align-items: flex-start;
+  justify-content: flex-start;
+`;
```
gustavs-gutmanis: What is the purpose of having it as `display: flex` and setting the item alignment to be that of a generic H3 element, and setting `width` and `height` to 100%, when all of that is out of the box for an H3 element?

We should strive to write as little repetitive code as possible without going nuts, of course, but this seems very redundant and introduces more problems than it solves.

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.styles.ts
```diff
@@ -0,0 +1,286 @@
...
+
+export const Row = styled.div`
+  margin: 0;
+  padding: 0;
+  width: 100%;
+  height: 100%;
+  display: flex;
+  flex-direction: row;
+  align-items: flex-start;
+  justify-content: flex-start;
```
gustavs-gutmanis: Margin and padding should be 0 here by default, and width will be 100% by default. Same goes for all the other components I guess.

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.styles.ts
```diff
@@ -0,0 +1,286 @@
...
+  padding: 0;
+  width: 100%;
+  height: 100%;
+  display: flex;
+  align-items: center;
+  flex-direction: column;
+  justify-content: center;
+`;
+
+export const Table = styled.table`
```
gustavs-gutmanis: Nowadays I think `display: grid` is a better tool for displaying tabular data, but if you're not into it, we can skip it for now.

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.styles.ts
```diff
@@ -0,0 +1,286 @@
...
+        border-bottom: 1px solid rgba(0, 0, 0, 0.1);
+
+        &:nth-last-child(1),
+        &:nth-last-child(2),
+        &:nth-last-child(3) {
+          padding: 6px 9px;
+          text-align: center;
+          width: 20px !important;
+          background-color: #ffffff;
+        }
```
gustavs-gutmanis: I would suggest using a classname for this purpose rather than hoping that the element order will always stay the same.

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.styles.ts
```diff
@@ -0,0 +1,286 @@
...
+  display: flex;
+  vertical-align: middle;
+
+  &::after {
+    margin: 0;
+    padding: 0;
+    content: '\\2715';
+    font-weight: 800;
+  }
+`;
```
gustavs-gutmanis: Lots of identical code that has only a thing or two differing.
You could create a base component and then extend them instead.

```suggestion
const BaseIcon = styled.div`
  display: flex;
  vertical-align: middle;
  
  margin: 1px 0 0 0;

  &::after {
    font-weight: 800;
  }
`;

export const DragIcon = styled(BaseIcon)`
  &::after {
    content: '\\2723';
  }
`;

export const DeleteIcon = styled(BaseIcon)`
  &::after {
    content: '\\2715';
  }
`;
```

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import { OptionsEditor } from './options-editor/options-editor';
+import { Control } from './control';
+import type { ControlType } from './types';
+
+const Options: React.FC<ControlType> = ({ field, property, dispatch }) => {
+  const { handle } = property;
+  const { uid, properties } = field;
+
+  property.label = 'Options Editor';
```
gustavs-gutmanis: This should be set in PHP in the `#[Property('Options Editor')` not overridden here in the code.

Otherwise we won't be able to call it whatever we want for some other place which might need different copy.

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+type PayloadProp = {
+  index: number;
+  key: string;
+  value: string | boolean;
+};
+
+export type CustomOptionsProps = {
+  useCustomValues: boolean;
+  options: OptionProp[] | GenericValue[];
```
gustavs-gutmanis: Why can it be both here?
I'm assuming you're entirely responsible for how the data structure ends up being, right?

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.tsx
```diff
@@ -0,0 +1,292 @@
...
+};
+
+type Props = {
+  handle: string;
+  value: CustomOptionsProps | GenericValue;
+  onChange?: (customOptions: CustomOptionsProps) => void;
```
gustavs-gutmanis: This will always be defined, so there's absolutely no necessity for complicating the codebase with `if (onChange)` calls.

```suggestion
  onChange: (customOptions: CustomOptionsProps) => void;
```

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.tsx
```diff
@@ -0,0 +1,292 @@
...
+      default: false,
+      autofocus: true,
+    });
+
+    if (onChange) {
+      onChange({
+        ...value,
+        options,
+      });
+    }
```
gustavs-gutmanis:
```suggestion
    onChange({
      ...value,
      options,
    });
```

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.tsx
```diff
@@ -0,0 +1,292 @@
...
+    options = options.filter(
+      (option: OptionProp, index: number) => index !== optionIndex
+    );
+
+    if (onChange) {
+      onChange({
+        ...value,
+        options,
+      });
+    }
```
gustavs-gutmanis:
```suggestion
```suggestion
    onChange({
      ...value,
      options,
    });
```
```

File: packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.tsx [hunk 12]
```diff
@@ -0,0 +1,292 @@
...
+  };
+
+  /**
+   * Filters out the option based on its index
+   * @param optionIndex
+   */
+  const deleteOption = (optionIndex: number): void => {
+    let options = JSON.parse(JSON.stringify(value.options));
```
gustavs-gutmanis: You should not be doing this `JSON.parse/stringify` hack. We must be able to rely on the original value.
And if we can rely on the original value, then you won't need to define the types of the `.filter()` arguments below, since they will be inferred from the original array.

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+    options = options.filter(
+      (option: OptionProp, index: number) => index !== optionIndex
```
gustavs-gutmanis:
```suggestion
      (option, index) => index !== optionIndex
```

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.tsx
```diff
@@ -0,0 +1,292 @@
...
+  };
+
+  const dragAndDropOption = (optionIndex: number): void => {
+    // TODO: Implement DnD once we have added DnD functionality elsewhere
+  };
+
+  /**
+   * Find and update option property value
+   * @param payload
+   */
```
gustavs-gutmanis: Lets not use doctype comments inside of typescript, since everything is already typed.

Also these kinds of comments are more harm than good
```ts
/**
 * Get Specific Item by index
 * @param number index
 */
function getSpecificItem(index: number) {
```

If we're doing or jobs right, the function name should describe what the function is doing, and the function should be doing only one thing.

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.tsx
```diff
@@ -0,0 +1,292 @@
...
+                <td>Label</td>
+                {value.useCustomValues && <td>Value</td>}
+                <td></td>
+                <td></td>
+                <td></td>
+              </tr>
+            </thead>
+            <tbody>
+              {value.options &&
+                value.options.map((option: OptionProp, index: number) => (
```
gustavs-gutmanis: I think you should be using `options` instead, which already has the correct type and a default value set up.

Also you can type it better and not have to type the `.map()` arguments.
```suggestion
              {options.map((option, index) => (
```

gustavs-gutmanis:
There is an issue in [AbstractExternalOptionsField.php:77](https://github.com/solspace/craft-freeform/blob/master/packages/plugin/src/Library/Composer/Components/Fields/AbstractExternalOptionsField.php#L57) which comes from the new changes not being supported by the actual field generator. 

Which is perfectly fine, just do a `return []` at the beginning of the method, and continue work, we will get to that part later.

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+    // Remove event listeners as cleanup
+    return () => {
+      window.removeEventListener('keyup', keyUpHandler);
+      window.removeEventListener('keydown', keyDownHandler);
+    };
+  }, []);
+
+  return keyPressed;
+};
```
gustavs-gutmanis: This should be a custom hook residing in its own file.

A test for this would also be nice, lets use [@testing-library:keyboard](https://testing-library.com/docs/user-event/keyboard) event tester. 
As well as the [@testing-library/react-hooks](https://react-hooks-testing-library.com/usage/basic-hooks) for testing hooks

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/options-editor.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { CustomOptions } from '@editor/builder/tabs/layout-editor/property-editor/form-controls/options-editor';
+import type { CustomOptionsProps } from '@editor/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options';
+
+import { Column, Heading, Row, Wrapper } from './options-editor.styles';
+
+type OptionsEditorProps = {
+  source: string;
```
gustavs-gutmanis: This should be an `enum` instead

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/options-editor.tsx
```diff
@@ -0,0 +1,58 @@
...
+  customOptions?: CustomOptionsProps;
+  // TODO - Implement other source types for Entries, Users, Predefined etc
+};
+
+type Props = {
+  handle: string;
+  value: OptionsEditorProps;
+  onChange?: (value: OptionsEditorProps) => void;
```
gustavs-gutmanis: Also no need to make it optional

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/options-editor.tsx
```diff
@@ -0,0 +1,58 @@
...
+                source: event.target.value,
+              })
+            }
+          >
+            <option value=""></option>
+            <option value="customOptions">Custom Options</option>
+          </select>
+        </Column>
+      </Row>
+      {value.source == 'customOptions' && (
```
gustavs-gutmanis: When we would have them as enums, we would check for it here using enums as well.

```ts
value.source === Source.CustomOptions
```

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/options-editor.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <CustomOptions
+          handle={handle}
+          value={(value.customOptions as CustomOptionsProps) || []}
```
gustavs-gutmanis: This should already be typed before it comes here.
`value` should reflect all of the values it will contain.
This will never change and is known to us beforehand.

gustavs-gutmanis (changes requested):
Great work, but leaving a handful of comments to look at.

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor/custom-options.tsx
seandelaney: @gustavs-gutmanis Typescript complains if I remove `OptionProp` and `number` here.

gustavs-gutmanis: Yes, because of `let options = JSON.parse(JSON.stringify(value.options));` the `options` variable is of type `any`, thus the `.filter` method cannot infer anything.

The solution should be typing the `value` property, so that typescript knows it contains `options` and that they are of type `OptionProp[]`. That would solve this problem

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor.styles.ts
```diff
@@ -0,0 +1,21 @@
...
+`;
+
+export const H3 = styled.h3`
+  width: 100%;
+  text-align: left;
+  margin: 0 0 10px 0;
+`;
+
+export const Row = styled.div`
+  width: 100%;
```
gustavs-gutmanis, line 14: I really doubt that `width: 100%` is needed here and for the others as well, width is automatically 100% for all block elements.

gustavs-gutmanis (approved):
Beautiful work @seandelaney 

I added two more files to split out the logic a bit for the custom checkboxes. Please take a look at it and let me know what you think.

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/options-editor.styles.ts
seandelaney: It's needed. I removed it and the layout broke on me.

PR #559, opened 2022-12-19
[SFT-23]: Property type: min/max

seandelaney:
Added new MinMax form control type component, that manages the min and max values of a number input field type.

Updated UI to inline both input fields on the same row instead of individual rows.


packages/plugin/src/Fields/NumberField.php
```diff
@@ -74,24 +64,33 @@ public function getValue(): mixed
         return $value;
     }
 
+    public function getMinMaxValues(): array
+    {
+        if (!\is_array($this->minMaxValues)) {
+            return [];
```
gustavs-gutmanis: lets return a `[null, null]` here, so that we can always rely on destructuring.
```php
[$min, $max] = $this->getMinMaxValues();
```

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/array.tsx
```diff
@@ -0,0 +1,37 @@
...
+              })
+            )
+          }
+        />
+      )}
+    </Control>
+  );
+};
+
+export default Array;
```
gustavs-gutmanis: Lets create a special field type `minMax` instead of having to do `if/else` inside a generic `array` field type. It will become unmanageable real fast this way.

And in PHP specify the new property type inside the `Property` attribute class. I'll set an example below.

packages/client/src/app/pages/forms/edit/builder/tabs/layout-editor/property-editor/form-controls/index.ts
```diff
@@ -1,3 +1,4 @@
+export { default as array } from './array';
```
gustavs-gutmanis: This is where the additional input type will come, instead of `array`, which might be a different thing entirely.

packages/plugin/src/Fields/NumberField.php
```diff
@@ -31,26 +31,16 @@ class NumberField extends TextField
...
     )]
     #[PropertyGroup('length')]
     protected ?int $maxLength = null;
 
     #[Property(
-        placeholder: 'Max',
-    )]
-    #[PropertyGroup(
-        'value',
         label: 'Min/Max Values',
```
gustavs-gutmanis, line 40:
```suggestion
        label: 'Min/Max Values',
        type: Property::TYPE_MIN_MAX,
```

gustavs-gutmanis: This is essentially going to be exactly like the `Options` property type, which handles all of its logic inside its own component.

packages/plugin/src/Fields/NumberField.php
```diff
@@ -74,24 +64,33 @@ public function getValue(): mixed
...
-    public function getMinLength(): int
+    public function getMinLength(): ?int
     {
         return $this->minLength;
     }
 
     public function getMinValue(): int
     {
-        return $this->minValue;
+        return $this->minMaxValues[0];
```
gustavs-gutmanis: This would die currently without having the values specified, since the default return value is `[]` instead of `[null, null]`

And maybe it would be prettier to do something like this:

```suggestion
        [$min] = $this->getMinMaxValues();

        return $min;
```

packages/plugin/src/Fields/NumberField.php
```diff
@@ -74,24 +64,33 @@ public function getValue(): mixed
...
     }
 
     public function getMaxValue(): int
     {
-        return $this->maxValue;
+        return $this->minMaxValues[1];
```
gustavs-gutmanis:
```suggestion
        [, $max] = $this->getMinMaxValues();

        return $min;
```

gustavs-gutmanis (commented):
Nice, added some comments

gustavs-gutmanis (approved):
Absolutely beautiful!

PR #565, opened 2022-12-27
[SFT-39]: Drag & Drop fields from field type section

gustavs-gutmanis:
* implemented dragging `FieldTypes` into new rows in the field layout


seandelaney (approved):
Very nice

PR #571, opened 2023-01-06
[SFT-135]: Property type: tabular data

seandelaney:
- Added new `table` field property control.
- Added new `TableLayoutEditor` component that allows a user to create tabular data by creating rows and columns (label, type and value).
- Added Property attributes to `TableField.php`
- Tidy up custom options folder for options field property control so it make more sense/ more descriptive.
- Fixed a minor typo.


gustavs-gutmanis (approved):
Amazing work @seandelaney!

PR #574, opened 2023-01-09
[SFT-40]: Drag & Drop fields from layout into other rows/columns

gustavs-gutmanis:
* adding drag & drop for cells
* adding drag & drop between rows
* adding drag & drop updated animations
* refactoring file structure

https://user-images.githubusercontent.com/709181/211390696-4612e02e-7f9e-4ab8-915b-494108f14be7.mov


seandelaney (approved):
G for Genius, after wrapping your head around these animations and making it work!!

PR #579, opened 2023-01-17
[SFT-34]: Pop-out functionality for more complex and wider properties

seandelaney:
- Added "preview" versions for Table and Options components.
- When said "previews" are clicked, "editor" versions of said components appear in a large container.
- Cleaned up Table and Options tabular styles/structure removing headings and add buttons.
- Added some basic animations.
- Added hooks to add new tabular rows upon pressing enter key.
- Added hook to close large container upon clicking outside of the large container.
- Some component cleanup - child components of Table and Options are no longer scattered about the place.

Notes:

- I could not get smoother/better animations/transitions like scale to work nicely.
- I could not get the large container to resize based on child component bounding box size. Still uses `40vw` with fixed height.
- When switching between field types in the sidebar, the overflow does not hide the hidden panel during transition (If that makes sense. You'll see what I mean once you pull and run these changes.)


gustavs-gutmanis:
* refactored the table editor component
* added a `<PreviewComponent>` which is capable of showing a block of preview data and handles all of the click/keyboard functionality for displaying a pop-up with editable data
* refactored the table editor
  * added keyboard shortcuts for navigating the table with arrow keys 
* refactored the table preview

seandelaney (commented):
@gustavs-gutmanis Lovin' the cleanup here!

gustavs-gutmanis (approved):
On behalf of @seandelaney I approve of this work.

PR #585, opened 2023-01-24
[SFT-35]: Favorite a field

gustavs-gutmanis:
* added a button for favoriting a field
* added an endpoint which handles receiving the info


seandelaney (approved):
Creates a snapshot - nice work!

PR #586, opened 2023-01-25
Fix an issue with custom table attributes firing on all element types

engram-design:
Hey crew, 

This seems to be incorrect, and throws an error with Formie's submissions - although it's also not great that this logic fires on **all** element indexes, not just `Submission` and `SpamSubmission` elements. It's only really an issue with Formie because we also use the prefix `form:`, most other element indexes won't be an issue.

This is because you're checking if the current element type is in an array `[Submission::class, SpamSubmission::class, true]` with `true` being always true, it'll always match. Looking at the rest of your code, I assume you're wanting to use the `strict` argument for `in_array`.

Cheers!


gustavs-gutmanis:
Hey @engram-design,

Thanks for catching this, it was definitely meant as the third param of `in_array` 😄

PR #587, opened 2023-01-27
[SFT-36]: Loading indicator for Save button

seandelaney:
- Added new PubSub topic for handing form processing
- Form processing state is changed when creating, updating or handling errors
- Hooked up for form processing state to the save button
  - Button label shows "Saving..." and is disabled when processing


gustavs-gutmanis:
Nice work!
One thing - the PubSub was introduced so that when the persist action is triggered, each logic block could subscribe to the persist event and attach their own data to the logic, and then react to whatever happens afterwards.

So I changed your code a bit to just dispatch the `processing: true` instead of having its own topic, because nothing will ever add anything to it, and everything can react to the redux store changes.

gustavs-gutmanis (approved):

PR #593, opened 2023-02-13
[SFT-59]: Value transformers for field/form properties

gustavs-gutmanis:
* added a property value transformer interface
* added `Table` transformer for table fields
* added `Options` transformer for all fields with options
* added tests for implemented transformers


seandelaney (approved):
Great changes, especially the slide-out animations!

PR #594, opened 2023-02-13
[SFT-222]: The error block (for required field) is loaded inside of the "opinion-scale-scales" UL for Opinion Scale fields, and causes the error to display incorrectly.

gustavs-gutmanis:
* added a new `data-field-root` attribute
* rendering errors in the field root if it is present


kjmartens (approved):
Looks good

PR #596, opened 2023-02-14
[SFT-46]: Add favorited fields to layout

gustavs-gutmanis:
* fetching favorite fields
* displaying favorite fields
* re-fetching favorite fields when new favorite added
* fav fields can be used in the layout


seandelaney (approved):
This is working very nice!

PR #601, opened 2023-02-20
[SFT-115]: Property type: attributes

gustavs-gutmanis:
* added a property type for `Attributes`
* added a preview and editor for `Attributes`
* refactored `Table` and `Options` property types to reduce user friction


seandelaney (approved):
Love it.

PR #603, opened 2023-02-23
[SFT-237]: Customer getting "Integrity constraint violation" error when submitting a form

gustavs-gutmanis:
* fixing an issue with the `Install.php` migration for Integration Queue table


kjmartens (approved):
Tested and works without issue

PR #607, opened 2023-02-27
[SFT-234]: Use #[Property] instead of BluePrint in integrations

gustavs-gutmanis:
* refactored integrations to use `#[Property]` attributes instead of `SettingBlueprint`
* refactored `OAuth` connector
* overall improved the integration flow
* restructured integrations to have their own folders


seandelaney (approved):
Very nice refactor. 

I think I can still use some of my work from SFT-196 though.

PR #609, opened 2023-02-28
[SFT-239]: When using Database Table session context, submitting a form gives this error: "Call to a member function getFormId() on null"

gustavs-gutmanis:
* fixed database context cleanup issue


kjmartens (approved):
Looks good

PR #615, opened 2023-03-02
[SFT-234]: Use #[Property] instead of BluePrint in integrations

gustavs-gutmanis:
* refactored mailing list integrations
* refactored payment integrations
* consolidated all integration listing, editing logic into one, rather than three controllers/services, etc.


seandelaney (approved):
Refactor working nicely!

PR #619, opened 2023-03-06
[SFT-32]: Persist integrations

gustavs-gutmanis:
* added logic for persisting forms
* persisting only changed values


seandelaney (approved):
All working nicely. Really like how we skip updates if no values change.

PR #620, opened 2023-03-07
[SFT-228]: "Email Template" dropdown options don't entirely respect the email settings

gustavs-gutmanis:
* fixed email templates displaying both groups even if just one selected


kjmartens (approved):
Seems delightful

PR #622, opened 2023-03-10
[SFT-196] - Introduce base field type controls

seandelaney:
- Cleaned up / centralised all the different form controls locations into one single location.
- Added new Label and Instruction components.
- Removed duplicated form controls such as text, text area, select, bool and int components that existed under the refactor folder.
- Updated the date picker and color picker components so they now use the same structure/property values as all other form controls.
- Updated Form Settings (the Behaviour & General tabs) so these now use the same set of form controls as Layout, Integrations etc.
- Updated favorite feature to use the text and label components from the form controls.
- Removed the lightswitch component, so we have a clean canvas to implement our context idea that will render components as slim or standard.


packages/client/src/app/components/form-controls/index.tsx
```diff
@@ -26,12 +28,13 @@ export const FormComponent: React.FC<Props> = ({
   property,
   context,
 }) => {
-  const FormControl = types[property.type];
+  const type = camelCase(property.type) as PropertyType;
```
gustavs-gutmanis, line 29: This seems like it might not work in all cases, why the `camelCase` here?

The `types[property.type]` specifically gets the correct component variable name, whereas `camelCase` might not be correct in certain situations.

packages/client/src/app/pages/forms/edit/builder/tabs/form-settings/field-component.tsx
```diff
@@ -0,0 +1,27 @@
...
+import type { Property } from '@ff-client/types/properties';
+
+import { useValueUpdateGenerator } from './use-value-update-generator';
+
+type Props = {
+  namespace: string;
+  property: Property;
+};
+
+export const FieldComponent: React.FC<Props> = ({ namespace, property }) => {
```
gustavs-gutmanis, line 14: Nice refactor!

gustavs-gutmanis (commented):
Overall great work, just had some comments

PR #626, opened 2023-03-13
[SFT-250]: `form_posted` cookie

gustavs-gutmanis:
* setting the cookie only if form limiting is enabled for the form


kjmartens (approved):
Tested, looks good 👍

PR #627, opened 2023-03-13
[SFT-114]: Refactor AbstractProperties remnants

gustavs-gutmanis:
* removed all `AbstractProperties` classes and updated code to use form settings instead
* implemented form Settings with `#[Property]` attributes
* refactored `Form` object namespace
* removed all `Composer` remnants
* moved around Field and Form interfaces and traits into better folder structure


seandelaney (approved):
Lovely cleanup. Tested and working as expected.

PR #628, opened 2023-03-15
[SFT-33]: Implement validators in builder based on fetched properties

gustavs-gutmanis:
* implemented validators
* showing errors in form settings
* showing errors in property editor
* displaying error state in tabs and pages


seandelaney (approved):
Nice work with the validation!

PR #632, opened 2023-03-17
[SFT-48] [SFT-129] - Notifications

seandelaney:
- Added/updated notifications tab, view, sidebar and property editor
- Added notifications routing, store, state, queries and API endpoints
- Started refactor of retrieving and persisting notifications
- Updated database structure with migration changes.

I did not add the `Add` button incase this conflicts with SFP-130

REINSTALL IS REQUIRED

```
INSERT INTO `freeform_notifications` (`id`, `name`, `handle`, `type`, `class`, `metadata`, `lastUpdate`, `dateCreated`, `dateUpdated`, `uid`) VALUES
(1, 'freeform5IntegrationTestAdmin', 'freeform5IntegrationTestAdmin', 'admin', 'Solspace\\Freeform\\Notifications\\Admin\\Admin', '{notificationTemplate: \"\"}', '2023-03-02 16:36:01', '2023-03-02 16:36:01', '2023-03-02 16:36:01', 'd4faac45-00bb-44f5-8536-0a2a3881e026');
```

```
INSERT INTO `freeform_forms_notifications` (`id`, `notificationId`, `formId`, `enabled`, `metadata`, `dateCreated`, `dateUpdated`, `uid`) VALUES
(1, 1, 1, 1, '{\"notificationTemplate\":\"TEST\"}', '2023-03-02 16:36:01', '2023-03-17 15:08:23', '85d1fec3-4734-aa43-051c-0a2a3881e026');
```


packages/plugin/src/Bundles/Notifications/Providers/FormNotificationsProvider.php
```diff
@@ -0,0 +1,54 @@
...
+    public function getForForm(?Form $form = null): array
+    {
+        $notifications = $this->notificationsService->getAllNotifications();
+
+        /** @var FormNotificationRecord[] $formNotificationRecords */
+        $formNotificationRecords = FormNotificationRecord::find()
+            ->where(['formId' => $form?->getId() ?? null])
+            ->indexBy('notificationId')
+            ->all()
+        ;
```
gustavs-gutmanis, line 35: We should rather return an empty array if the form is null, since we need a form here, right?
The can't be a form notification without an actual form.

packages/plugin/src/Bundles/Notifications/Providers/NotificationDTOProvider.php
```diff
@@ -0,0 +1,118 @@
...
+        $type = $type ? $type->newInstance() : null;
+
+        /** @var Type $type */
+        if (!$type) {
+            return null;
+        }
+
+        $icon = $type->iconPath;
+        if ($icon) {
+            [$_, $icon] = \Craft::$app->assetManager->publish($icon);
```
gustavs-gutmanis, line 100: I think it makes more sense to have the icon assigned to the form category, rather than each notification template.

It makes sense for integrations because they display the icon of the service provider, so it can change from integration to integration in the CRM category, for instance, but these will never change between admin notifications.

packages/plugin/src/Bundles/Persistence/NotificationPersistence.php
```diff
@@ -0,0 +1,76 @@
...
+                continue;
+            }
+
+            $record->enabled = (bool) $enabled;
+            $record->metadata = json_encode((object) array_merge($metadata, $values));
+
+            $record->save();
+
+            if ($record->hasErrors()) {
+                $errors[$record->notificationId] = $record->getErrors();
```
gustavs-gutmanis, line 68: We should revisit this later, because the errors will be per field and they will be validated with `#[PropertyValidator]` instead of the Yii2 Record validators.
This will require some extra work is what I'm saying.
Maybe a `TODO` here would be nice

packages/plugin/src/Events/Notifications/NotificationResponseEvent.php
```diff
@@ -0,0 +1,42 @@
...
+
+    private ResponseInterface $response;
+
+    public function __construct(AbstractNotification $notification, ResponseInterface $response)
+    {
+        $this->notification = $notification;
+        $this->response = $response;
+
+        parent::__construct();
+    }
```
gustavs-gutmanis, line 21: With [PHP8 there's a neat little improvement](https://stitcher.io/blog/constructor-promotion-in-php-8) where you can declare the variables and their visibility directly in the constructor

```suggestion
    public function __construct(
        private AbstractNotification $notification, 
        private ResponseInterface $response
    ) {
        parent::__construct();
    }
```

packages/plugin/src/Library/DataObjects/Notifications/NotificationSetting.php
gustavs-gutmanis, line 1: This seems like it is unused anywhere and can be deleted

packages/plugin/src/Library/Notifications/AbstractNotification.php
gustavs-gutmanis, line 1: We don't need most of the handlers and translators here.
And we won't be doing the custom field value transformation here in this class, that was a bad pattern I used a long time ago.

Just have it have the properties and name, handle, id and the property processing. the rest should be removed.

packages/plugin/src/Library/Notifications/DataObjects/FieldObject.php
gustavs-gutmanis, line 1: What is this for, I wonder? Doesn't seem like we will need something like this.
Is this for the field mapping?

It seems like something from the old integrations. We'll be using something better from now on.

packages/plugin/src/Library/Notifications/NotificationInterface.php
```diff
@@ -0,0 +1,32 @@
...
+
+interface NotificationInterface
+{
+    public const FLAG_GLOBAL_PROPERTY = 'global-property';
+
+    public const FLAG_INTERNAL = 'internal';
+
+    public const FLAG_ENCRYPTED = 'encrypted';
+
+    public const FLAG_READONLY = 'readonly';
```
gustavs-gutmanis, line 23: This is the same as in integrations, and if we're reusing those, we should just move them to somewhere that makes sense for all to use rather than just copy them over.

packages/plugin/src/Library/Notifications/OAuth/RefreshTokenInterface.php
gustavs-gutmanis, line 1: Email notifications aren't really like integrations, they will never connect to an outside source, so OAuth is definitely out of scope here

packages/plugin/src/Models/NotificationModel.php
```diff
@@ -0,0 +1,113 @@
...
+            $this->handle ?? '',
+            $this->name ?? '',
+            $this->lastUpdate,
+            $this->metadata,
+            FreeformLogger::getInstance($logCategory),
+            new CraftTranslator(),
+            $handler,
+            \Craft::$container->get(PropertyProvider::class),
+        );
+    }
```
gustavs-gutmanis, line 104: We don't need the handlers here and translators, this is again an outdated old pattern, which blows. Lets remove this, and I'll take care of it when I work on Dynamic Notifs

packages/plugin/src/Notifications/BaseNotification.php
gustavs-gutmanis, line 1: Love the idea of each template type defined as a class with its own properties and everything, very nice work!

packages/plugin/src/Services/NotificationsService.php
```diff
@@ -27,89 +30,256 @@ class NotificationsService extends BaseService
...
+            throw $exception;
+        }
     }
 
-    public function create(string $name): NotificationRecord
+    /**
+     * @throws \yii\base\Exception
+     * @throws \yii\base\InvalidConfigException
+     */
+    public function decryptModelValues(NotificationModel $model)
```
gustavs-gutmanis, line 181: We won't be doing what we're doing with integrations, so there won't be no `internal` and `encrypted` values, etc

gustavs-gutmanis (changes requested):
Great work!
Just need some touch-ups before merging

gustavs-gutmanis (approved):

PR #654, opened 2023-04-14
[SFT-130]: Notifications: Dynamic Notifications

gustavs-gutmanis:
* refactored `FieldInterface` to use `Attributes` instead of the old custom properties
* refactored pop-ups in builder
* refactored keyboard navigation in tables
* implemented dynamic notifications editor
* implemented notification persistence


seandelaney (approved):
What I tested worked as expected. I can see notification data I input saved in the DB.

Nice work.

PR #656, opened 2023-04-18
[SFT-280] - Include template autosuggest to relevant Freeform path settings in CP

seandelaney:
- Replaced text field with auto suggest field for email, formatting and success template paths
- Added helper function to list all directories within craft site templates folder


gustavs-gutmanis (approved):
Looks great!

PR #657, opened 2023-04-18
[SFT-118]: create Rules sidebar with mini-map

gustavs-gutmanis:
* refactored store structure
* added rules mini-map


kjmartens (approved):
Tasty!

PR #660, opened 2023-04-20
[SFT-304] - Add the ability to parse Twig inside the Preset Assets feature on email notifications

seandelaney:
- Added new new helper function that detects if a value is using Twig tags.
- Updated `EmailTemplate` so that if the Preset Assets value is using Twig tags, its passed as is. E.g not comma separated.
- Updated `NotificationRecord` so that if the Preset Assets value is using Twig tags, the value is returns as non JSON decoded.
- Updated `NotificationFileService` to correctly save the Preset Assets value, regardless of whether it's using Twig tags or a separated comma list value.
- Updated notifications `edit` template so the Preset Assets field is hidden if the template is File based and the Preset Assets value is using Twig tags.
- Updated `MailerService` to detect, parse and convert value to correct format if the Preset Assets field is using Twig tags.

Tested using

```
{# presetAssets:  1442, 1443 #}
{# presetAssets:  {{ '1442, 1443' }} #}
{# presetAssets:  {{ freeformGlobals.field1 }} #}
{# presetAssets:  {{ freeformGlobals.field3.ids|join(',') }} #}
```


packages/plugin/src/templates/notifications/edit.twig
gustavs-gutmanis, line 1: I think it's best to do all of this logic inside the `Edit` action of the notification controller, rather than twig.

gustavs-gutmanis (changes requested):

gustavs-gutmanis (approved):
Nice!

PR #665, opened 2023-04-25
[SFT-119]: rule editor for active field

gustavs-gutmanis:
* added mini-map for rule navigation
* added field rules
  * value input varies from operator to field type
* adding reactive mini-map to display rule effects


seandelaney (approved):
Adding rules to fields is working nicely. Visual updates by use of icons also very cool!

PR #670, opened 2023-04-28
[SFT-122]: create a database table, records and service for storing and fetching rules

gustavs-gutmanis:
* added base record for rules
* added child record for field rules
* added record for conditionals
* fetching rules data via endpoint
* displaying loaders
* persisting/loading of field rules implemented


seandelaney (approved):
Working nicely.

PR #676, opened 2023-05-10
[SFT-133]: Conditional notifications

gustavs-gutmanis:
* added conditional notifications type
* added editor for it
* added a conditional rule field editor
* refactored saving of rules and notifications (only saves when accessed at least once)


seandelaney (approved):
Very nice. Tested and all working

PR #677, opened 2023-05-10
[SFT-264] - GraphQL Mutations - Submissions

seandelaney:


packages/plugin/src/Bundles/GraphQL/Arguments/Inputs/RecaptchaInputArguments.php
```diff
@@ -0,0 +1,52 @@
...
+
+        // or if the form has payment fields, then bail
+        if (\count(self::$form->getLayout()->getFields(PaymentInterface::class))) {
+            return [];
+        }
+
+        $fields = [];
+
+        foreach (self::$form->getLayout()->getFields() as $field) {
+            if ($field instanceof RecaptchaField) {
```
gustavs-gutmanis: You could do this to narrow the fields down:

```suggestion
        $fields = self::$form->getLayout()->getFields(RecaptchaField::class);
        foreach ($fields as $field) {
```

packages/plugin/src/Bundles/GraphQL/Interfaces/FormInterface.php
```diff
@@ -153,33 +156,83 @@ public static function getFieldDefinitions(): array
...
+                        if ($source->getPropertyBag()->get(Form::DATA_DISABLE_RECAPTCHA)) {
+                            return null;
+                        }
+
+                        // or if the form has payment fields, then bail
+                        if (\count($source->getLayout()->getFields(PaymentInterface::class))) {
+                            return null;
+                        }
+
+                        foreach ($source->getLayout()->getFields() as $field) {
```
gustavs-gutmanis: Same as above, you can narrow it down directly.

packages/plugin/src/Fields/FileUploadField.php
```diff
@@ -129,7 +138,16 @@ public function getInputHtml(): string
     public function uploadFile()
     {
         if (!isset(self::$filesUploaded[$this->handle])) {
-            $response = $this->getForm()->getFileUploadHandler()->uploadFile($this, $this->getForm());
+            $response = null;
+
+            if ($this->getForm()->isGraphQLPosted()) {
+                try {
+                    $response = $this->getForm()->getFileUploadHandler()->uploadGraphQL($this, $this->getForm());
+                } catch (GuzzleException|InvalidSubpathException|Exception $e) {
```
gustavs-gutmanis:
```suggestion
                } catch (Exception $e) {
```
`Exception` covers all of the cases.

packages/plugin/src/Fields/FileUploadField.php
```diff
@@ -129,7 +138,16 @@ public function getInputHtml(): string
...
+                }
+            } else {
+                $response = $this->getForm()->getFileUploadHandler()->uploadFile($this, $this->getForm());
```
gustavs-gutmanis: It kind of feels wrong to have this GQL check logic inside the field, rather than the fileUploadHandler.

The `::uploadFile()` method receives `Form` and should do the check inside itself.

gustavs-gutmanis (commented):
A couple of comments, otherwise - tremendous work

gustavs-gutmanis (approved):
Tested, works and looks great!

gustavs-gutmanis (approved):
Tested, works and looks great!

gustavs-gutmanis (approved):
Thank you for your service 🦾

PR #680, opened 2023-05-15
[SFT-325] Field previews in builder layout

gustavs-gutmanis:
* adding field previews in builder layout
* using JS templates for displaying the previews
* using a skinnier input display in property editor (not showing instructions)
* instructions in property editor now on label hover


seandelaney (approved):
Really liking this preview work. Working nicely in testing.

PR #682, opened 2023-05-17
[SFT-329]: Add properties and preview for Opinion Scale Field

gustavs-gutmanis:
* added `opinion-scale` field logic
* added properties to property editor
* added live preview template
* refactored `ControlType` properties


seandelaney (approved):
Lovely

PR #683, opened 2023-05-18
[SFT-330]: Do not allow negative numbers in Textarea field row configuration

gustavs-gutmanis:
* added `min` and `max` configuration items to `Input\Integer` properties


seandelaney (approved):
Works great

Issue #684, opened 2023-05-18
Error due to missing submissionId after submitting a form


lbriet:
**Describe the bug or issue you're experiencing**
- Submitting a form (successfully i.e. no spam) always triggers a TypeError:
```
Solspace\Freeform\Services\Pro\Payments\PaymentsService::getBySubmissionId(): Argument #1 ($submissionId) must be of type int, null given, called in /var/www/html/vendor/solspace/craft-freeform/packages/plugin/src/Services/SubmissionsService.php on line 265
```
![Screenshot 2023-05-18 at 17 34 03](https://github.com/solspace/craft-freeform/assets/16916427/b2882766-9b87-445c-9f6b-5d7bcd86b227)

**Steps to reproduce**

1. Submit form

**Expected behavior**

1. Redirect to success template/page

**Craft & Plugin Info (please complete the following information):**

- Craft Version: 4.4.11
- Freeform Version: 4.0.24
- Freeform Edition: Pro
- Fresh Install or Upgrade: Upgrade

**Additional context**
- Downgrading to 4.0.23 returns all to normal
- Seeing the 4.0.24 changelog: we don't make use of any Payment gateways, nor have we ever configured one
- We don't save submissions to database
- Issue lies in https://github.com/solspace/craft-freeform/commit/2e1c43a39e6fe6bca5d03b65bcd1e2f3b1d03998#diff-5171b8575630cee7b531ee8f53f582c21bc037bcd7aa74740b1af43541f88a28 ?

kjmartens:
I'm very sorry for the trouble @lbriet. I can duplicate this in my testing as well. It seems to only happen if NOT saving submission data for the form. We'll make a fix for it shortly. 🙂

kjmartens:
This has been resolved in Freeform 4.0.24.1. 🙂

lbriet:
Thank you for the lightning quick fix!

PR #685, opened 2023-05-18
[SFT-326] - Missing submissionId error when submitting a form that does not store data #684

seandelaney:
Checks for submission ID before attempting to check for payment info


kjmartens:
Looks good 🙂

PR #686, opened 2023-05-19
[SFT-132]: Email field notifications

gustavs-gutmanis:
* adding email field notifications
* introduced `implements` filter to `Input\Field` which forces the property in react to show only fields which implement specific interfaces.


seandelaney (approved):
Nice work. Tested and does its job!

PR #689, opened 2023-05-31
[SFT-337]: Form rendering in twig

gustavs-gutmanis:
* refactored a lot of files to make form rendering possible at all
* added the initial work for better formatting templates
* added work that will aid in `Attributes` refactor


seandelaney (approved):
Looks good

PR #692, opened 2023-06-01
[SFT-47]: Add another form's field to layout

seandelaney:
- Added new endpoint that returns all forms fields.
- Updated a forms layout sidebar to show other forms field types.
- Allow users to add other forms fields to current form.


gustavs-gutmanis (approved):
Looks great

PR #693, opened 2023-06-02
[SFT-344]: Can't add Freeform Field Value widget to Dashboard

gustavs-gutmanis:
* fixed field value widget crashing on create
* it used the legacy submissions table to check for values


kjmartens (approved):
Looks good :)

PR #694, opened 2023-06-02
[SFT-338]: Refactor attributes into `Array<{ key: string, value?: string }>` 

gustavs-gutmanis:
* attributes are now storing dictionaries of data instead of arrays of [key, value]
* nested attributes receive data from merge automatically
* added table attributes for table fields letting you edit table specific attributes via builder


seandelaney (approved):
Looks great!

PR #695, opened 2023-06-06
[SFT-342]: formatting template refactor

gustavs-gutmanis:
* refactored formatting templates to use attributes
* builder select fields now accept nested options for `<optgroup />`
* refactored events attaching attributes
* added formatting template options to builder settings
* layouts now contain all field references, each row contains its field references, page does too, etc. (lets you access directly related fields per object scope)
* improved attribute cleanup when rendering them as strings (added tests for it too)


seandelaney (approved):
Nice work. Some really nice cleanup here too!

PR #696, opened 2023-06-07
[SFT-339]: Remove `Field::hash` as it is no longer used

gustavs-gutmanis:
* removed `Field::hash()` calls


seandelaney (approved):
Nice work. 

FYI, I'm glad I moved away from using Recaptcha field handle in our GraphQL mutations branch y'day as it would have caused breaking changing.

PR #698, opened 2023-06-08
[SFT-343]: JS Honeypot and multi-page payload issue

gustavs-gutmanis:
* fixed payload multi-page forms from submitting when triggering an invalid submission with ajax when JS honeypot is enabled


kjmartens (approved):
Looks good

PR #702, opened 2023-06-13
SFT-349 - Port over FF4 GraphQL functionality

seandelaney:


packages/plugin/src/Fields/AbstractField.php
```diff
@@ -133,7 +135,16 @@ public function getValue(): mixed
...
+        if ($this instanceof RatingField) {
+            // Prevents GraphQL from triggering an number constraint violation
+            if (!empty($this->getValue())) {
+                $this->value = $value;
+            } else {
+                $this->value = null;
+            }
+        } else {
+            $this->value = $value;
+        }
```
gustavs-gutmanis: Now each field takes care of its own logic if it differs from the base functionality.
So you should override the `setValue()` in `RatingField` itself.

packages/plugin/src/Fields/AbstractField.php
```diff
@@ -553,7 +598,9 @@ protected function onAfterInputHtml(): string
      */
     protected function validate(): array
     {
-        $event = new ValidateEvent($this);
+        $form = $this->getForm();
+
+        $event = new ValidateEvent($this, $form);
```
gustavs-gutmanis: What's the benefit of having the `$form` variable here if it can be extracted from the field?

At the very least we could extract it out of the field inside the `ValidateEvent` constructor and then add `implements FormEventInterface` as well.

packages/plugin/src/Fields/Implementations/FileUploadField.php
```diff
@@ -170,108 +174,257 @@ public function uploadFile(): ?array
         return self::$filesUploaded[$this->handle];
     }
 
+    public function getContentGqlType(): array|\GraphQL\Type\Definition\Type
```
gustavs-gutmanis: Lets use short imports, if the name is used already, alias the import to something like `GQLType`

packages/plugin/src/Fields/Implementations/HtmlField.php
```diff
@@ -74,4 +74,9 @@ public function getInputHtml(): string
 
         return $this->getValue();
     }
+
+    public function includeInGqlSchema(): bool
```
gustavs-gutmanis, line 78: We should have all the GQL related methods that were added to `AbstractField` also added to the `FieldInterface` I guess.

packages/plugin/src/Fields/Implementations/Pro/OpinionScaleField.php
```diff
@@ -101,6 +101,40 @@ public function getLegends(): TabularData
         return $this->legends;
     }
 
+    public function getContentGqlMutationArgumentType(): array|\GraphQL\Type\Definition\Type
```
gustavs-gutmanis: Lets import the `Type` class to use the short name, please

packages/plugin/src/Fields/Implementations/Pro/TableField.php
```diff
@@ -184,6 +186,65 @@ public function setValue(mixed $value): self
...
+    }
+
+    public function getContentGqlMutationArgumentType(): array|\GraphQL\Type\Definition\Type
+    {
+        $layout = [];
+        $textValuesInclude = '';
+        $selectValuesInclude = '';
+        $checkboxValuesInclude = '';
+
+        foreach ($this->getTableLayout() as $column) {
```
gustavs-gutmanis, line 202: The way table layout is used changed in v5, please see if it works with the new approach. You can check how it's used in the render method.

packages/plugin/src/Form/Form.php
```diff
@@ -121,6 +124,10 @@ abstract class Form implements FormTypeInterface, \IteratorAggregate, \Countable
     private bool $pagePosted = false;
     private bool $formPosted = false;
 
+    private bool $graphqlPosted = false;
+
+    private array $graphqlArguments = [];
```
gustavs-gutmanis, line 129: What are these I wonder? The posted GQL values?
Would it be possible to have them outside of the `Form` instance?

There's lots of stuff we can eventually move outside of the Form itself, it should be as barebones as possible.

If it's not possible then forget about it.

packages/plugin/src/Services/SubmissionsService.php
```diff
@@ -142,6 +144,42 @@ public function getSubmissionCountByForm(bool $isSpam = false, Carbon $rangeStar
...
+     * @throws FreeformException
+     */
+    public function handleSubmission(Form $form, Submission $submission): void
+    {
+        $freeform = Freeform::getInstance();
+
+        $formsService = $freeform->forms;
+        $spamSubmissionsService = $freeform->spamSubmissions;
+
+        $event = new \Solspace\Freeform\Events\Forms\SubmitEvent($form, $submission);
```
gustavs-gutmanis: Lets always import the FQN and use the import names instead.

packages/plugin/src/controllers/SubmitController.php
```diff
@@ -63,8 +71,10 @@ public function actionIndex(): ?Response
                 return $this->toAjaxResponse($form, $submission, $returnUrl);
             }
 
-            if (BehaviorSettings::SUCCESS_BEHAVIOUR_LOAD_SUCCESS_TEMPLATE === $form->getSuccessBehaviour()) {
-                if ($form->getSuccessTemplate()) {
+            $behavior = $form->getSettings()->getBehavior();
+
+            if (BehaviorSettings::SUCCESS_BEHAVIOUR_LOAD_SUCCESS_TEMPLATE === $behavior->successBehavior) {
+                if ($behavior->successTemplate) {
```
gustavs-gutmanis, line 77: nice

gustavs-gutmanis (changes requested):
Nice and tedious work there, left a couple of comments here and there.

packages/plugin/src/Fields/AbstractField.php
seandelaney: Yep, good to know. I will update. 

I purposely didn't tag you for review yet, as I was still testing/checking over files some more to make sure I carried over all changes but wanted to publish the branch so I could see a clearer change log.

packages/plugin/src/Form/Form.php
seandelaney: As per Slack chat, we've agreed to continue focus on existing working functionality. Refactoring/improvements to be investigated/applied in the future.

gustavs-gutmanis (approved):
Gorgeous

PR #705, opened 2023-06-15
[SFT-375] - Add `form` as a reserved keyword for field handles

seandelaney:


seandelaney:
Closing as this commit will be included within SFT-356

PR #706, opened 2023-06-19
[SFT-412] - The automatic Submission Purge feature removes submissions, but does not seem to delete associated Asset files

seandelaney:


kjmartens (approved):
Tested, works.

PR #708, opened 2023-06-20
[SFT-358] - Freeform incorrectly calls craft\services\Assets::getFold…

seandelaney:
…erById() when dynamically creating a directory for file uploads


kjmartens (approved):
Looks good

PR #710, opened 2023-06-22
[SFT-412] - The automatic Submission Purge feature removes submissions, but does not seem to delete associated Asset files (#706)

seandelaney:
- Added `skipContent(true)` flag back in as it was causing weird/random/missing submissions in the result set.
- I originally removed `skipContent(true)` flag as it was causing all submission field values to be null in the result set, meaning the asset ids array was always empty. Hence why assets were never deleted during purge of submissions.
- We now grab fresh submission instance that now contains all field values, allowing use to grab asset ids and correctly delete assets during purging.

### DIFF
--- a/packages/plugin/src/Services/SubmissionsService.php
+++ b/packages/plugin/src/Services/SubmissionsService.php
@@ -512,6 +512,7 @@ class SubmissionsService extends BaseService implements SubmissionHandlerInterfa
 
         $query = $this->getFindQuery()
             ->id($ids)
+            ->skipContent(true)
         ;
 
         $count = $query->count();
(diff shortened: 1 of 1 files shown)


packages/plugin/src/Services/SubmissionsService.php
```diff
@@ -523,6 +524,7 @@ public function purgeSubmissions(int $age = null): array
         foreach ($query->batch() as $results) {
             /** @var Submission $submission */
             foreach ($results as $submission) {
+                $submission = $this->getSubmission($submission->getId());
```
gustavs-gutmanis, line 527: Does this make a single query per result?
This would make it extremely slow in large datasets.
Maybe it's possible to extract all of the IDs first and fetch the submissions in a single query? 
Though that might also cause issues... not sure how to best approach this.

PR #712, opened 2023-06-22
[SFT-456] - When reCAPTCHA is enabled, the Drag and Drop File Upload …

seandelaney:
…field no longer shows existing file(s) previews when page is reloaded (due to going through pages, etc)

Forms with recaptcha enabled, have multiple pages with a File DnD field would send `g-recaptcha-response` with a blank value when grabbing existing files when switching between pages. The latest changes to recaptcha couldn't handle the blank value.


kjmartens (approved):
Looks good now

PR #714, opened 2023-06-23
Fix dutch translation

IbeGarritsen:
Fixed a typo in the dutch translation file


kjmartens:
Thanks! We'll include it in the next update 🙂

PR #717, opened 2023-06-27
[SFT-347]: Move Submit / Previous / Save buttons to be built-in

gustavs-gutmanis:
* Moved submit/back/save buttons to be built into pages
* Created admin edit panel for them
* Added a frontend renderer for the buttons that listens to `before closing tag` event and adds markup via that event
* added clearer animations when switching property editor focus to different fields


packages/plugin/src/Bundles/Form/SubmitButtons/SubmitButtons.php
```diff
@@ -0,0 +1,55 @@
...
+    public function __construct()
+    {
+        Event::on(
+            Form::class,
+            Form::EVENT_RENDER_BEFORE_CLOSING_TAG,
+            [$this, 'addSubmitButtons']
+        );
+    }
+
+    public function addSubmitButtons(RenderTagEvent $event): void
```
seandelaney: My OCD says no. How about just `add` ?

gustavs-gutmanis: sounds better

seandelaney (approved):
Nice work

PR #726, opened 2023-07-07
[SFT-140]: Element Connections: add field mapping logic

gustavs-gutmanis:
* added field mapping controls to builder


seandelaney (approved):
Works well.

PR #732, opened 2023-07-12
[SFT-421] - Fixed CP translations

seandelaney:


kjmartens (approved):
This likely isn't used anymore, but this does fix an issue 😄

PR #743, opened 2023-07-20
[SFT-482] - Expose form settings for GraphQL query

seandelaney:
- Exposes more form settings/properties via GraphQL
- Fixed ReCaptcha bug when querying forms via GraphQL


kjmartens (approved):
Seems to work good

PR #744, opened 2023-07-20
[SFT-458]: Form Builder Cleanup

gustavs-gutmanis:
* conditional rules for frontend
* page deletion
* field deletion
* default page when creating form
* field type changer
* renamed several field types
* automatic handle generator


seandelaney (approved):
Works nicely. Love the validation extraction into bundles.

PR #748, opened 2023-07-22
Porting FF4 features and bug fixes

seandelaney:


packages/client/src/app/pages/forms/edit/store/slices/fields/index.ts
```diff
@@ -77,7 +78,12 @@ export const fieldsSlice = createSlice({
       );
     },
     edit: (state, action: PayloadAction<EditType>) => {
-      const { uid, handle, value } = action.payload;
+      let { value } = action.payload;
+      const { uid, handle } = action.payload;
+
+      if (handle == 'handle') {
+        value = transliterate(value);
+      }
```
gustavs-gutmanis: We should move this to the `handle` middleware, rather than have a hardcoded specific case in the store slice

packages/client/src/app/pages/forms/edit/store/slices/form/index.ts
```diff
@@ -47,12 +48,17 @@ export const formSlice = createSlice({
...
+      let { value } = payload;
+      const { namespace, key } = payload;
 
       if (!state.settings[namespace]) {
         state.settings[namespace] = {};
       }
 
+      if (key == 'handle') {
+        value = transliterate(value);
+      }
```
gustavs-gutmanis: Moving this to `handle` middleware will take care of all places at once

packages/plugin/src/Bundles/Form/Limiting/FormLimiting.php
```diff
@@ -60,38 +70,109 @@ public function handleLimitations(ValidationEvent $event): void
...
+
+        // Get all email fields on the form
+        $emailFields = $form->getLayout()->getFields(EmailField::class);
+
+        // Get all email field values
+        $emailFieldValues = [];
+        foreach ($emailFields as $emailField) {
+            if ($request->getIsCpRequest() || $request->getIsConsoleRequest()) {
+                continue;
+            }
```
gustavs-gutmanis: Shouldn't this line be outside of the loop, at the very top?

packages/plugin/src/Bundles/Form/Limiting/FormLimiting.php
```diff
@@ -60,38 +70,109 @@ public function handleLimitations(ValidationEvent $event): void
...
+
+        $isPosted = (bool) $query->scalar();
+
+        if ($isPosted) {
+            $this->addMessage($form, $event);
         }
     }
 
-    private function limitByCookie(Form $form): void
+    private function limitByCookie(Form $form, FormEventInterface $event): void
```
gustavs-gutmanis: Now that I think about it, looks like passing `$form` as an argument here is kind of redundant, since the `$event` object contains the form

packages/plugin/src/Freeform.php
```diff
@@ -249,7 +247,7 @@ public function init(): void
         $this->initServices();
         $this->initIntegrations();
         $this->initTwigVariables();
-        $this->initWidgets();
+        // $this->initWidgets();
```
gustavs-gutmanis: This should be removed, if it's moved to Feature Bundle.

gustavs-gutmanis (commented):
Nice work, leaving a couple of comments which might lead somewhere

packages/client/src/app/pages/forms/edit/store/slices/fields/index.ts
seandelaney: Yes, agreed. I did think about middleware but couldn't find any working examples. I was going to chase you on this but forgot. I will update.

packages/plugin/src/Bundles/Form/Limiting/FormLimiting.php
seandelaney: This will make for a nice improvement.

packages/client/src/app/components/middleware/implementations/handle.test.ts
```diff
@@ -16,4 +16,15 @@ describe('handle middleware', () => {
 
     expect(result).toEqual('this_is-underscored');
   });
+
+  it('converts unicode characters to Latin characters', () => {
+    const string =
+      'Visi cilvēki piedzimst brīvi un vienlīdzīgi savā pašcieņā un tiesībās';
```
gustavs-gutmanis, line 22: 🤣

gustavs-gutmanis (approved):
Nice!

packages/client/src/app/components/middleware/implementations/handle.test.ts
seandelaney: Least you understood...

PR #750, opened 2023-07-26
[SFT-489]: Update integrations to use combined approach

gustavs-gutmanis:
* refactoring integrations to use field mappings
* refactoring integrations to use less events
* adding `#[EventListener]` attributes that can be used to add event listeners to integrations
* adding style updates for integrations in builder


seandelaney (approved):
All good.

PR #754, opened 2023-07-31
[SFT-487] - Form Builder Settings - Mainly Visibility Filters

seandelaney:
- Added Visibility Filters to Settings tab fields.
- Updated Property attribute with new width property to control layout.


packages/client/src/app/pages/forms/edit/store/slices/form/form.selectors.ts
```diff
@@ -7,13 +7,23 @@ export const formSelectors = {
...
     all:
-      (namespace: string) =>
+      () =>
       (state: RootState): SettingsNamespace =>
-        state.form.settings?.[namespace] || {},
+        state.form.settings || {},
     one:
-      (namespace: string, key: string) =>
+      (namespace: string) =>
       (state: RootState): any =>
```
gustavs-gutmanis: Lets use the `GenericValue` that we have set, otherwise we get `any` related warnings. The `GenericValue` is an ignored `any` value.

gustavs-gutmanis (approved):
Looking great, just a tiny update on the `any` values and we're golden!

PR #756, opened 2023-08-03
[SFT-492] - Update Pipedrive CRM Field Mapping

seandelaney:
I've disabled animation as some fields do not show correctly with visibility filter.

Salesforce updates to support Custom URL's ~fails to authorise. Saving does work~.


packages/plugin/src/Integrations/CRM/Pipedrive/BasePipedriveIntegration.php
```diff
@@ -81,45 +65,31 @@ public function checkConnection(): bool
...
-     * Perform anything necessary before this integration is saved.
-     */
-    public function onBeforeSave()
+    public function fetchFields(string $category): array
     {
-        $client = $this->generateAuthorizedClient();
+        $endpoint = strtolower($category).'Fields';
+
+        if ('Lead' === $category) {
+            $endpoint = 'dealFields';
```
gustavs-gutmanis: I'm guessing Leads use the same fields as Deals here, right?

gustavs-gutmanis (approved):
Great work!

packages/plugin/src/Integrations/CRM/Pipedrive/BasePipedriveIntegration.php
seandelaney: Yes

PR #757, opened 2023-08-04
[SFT-493] - Update Zoho CRM Field Mapping

seandelaney:
Based off SFT-492

Zoho is currently using API v2 - my focus was to get the integration working using the current API version. So I structured the integration in a way where I can add in v5 without breaking v2.


gustavs-gutmanis (approved):
Looking great

PR #760, opened 2023-08-07
[SFT-515] - Update Pardot CRM Field Mapping

seandelaney:


seandelaney:
Merged into https://github.com/solspace/craft-freeform/pull/765

PR #761, opened 2023-08-07
[SFT-494] - Update HubSpot CRM Field Mapping

seandelaney:


seandelaney:
Merged into [SFT-511 ](https://github.com/solspace/craft-freeform/pull/765)

PR #762, opened 2023-08-08
[SFT-502] Update the Property editor to disappear when you click outside of field layout/fields area

gustavs-gutmanis:
* adding `useClickOutside` hook to props editor
* adding extra functionality to the hook which lets you supply an extra list of classes to exclude more elements from triggering the click


seandelaney (approved):
Very nice

PR #765, opened 2023-08-09
[SFT-511] - Update ActiveCampaign CRM

seandelaney:
Includes Pardot and HubSpot CRM too


seandelaney:
Included in https://github.com/solspace/craft-freeform/pull/766

PR #766, opened 2023-08-09
[SFT-514] - Update Insightly CRM

seandelaney:


packages/plugin/src/Integrations/CRM/ActiveCampaign/BaseActiveCampaignIntegration.php
```diff
@@ -0,0 +1,572 @@
...
+            throw new IntegrationException('Could not fetch fields for Contact');
+        }
+
+        $fieldList = [];
+
+        $fieldList[] = new FieldObject(
+            'listId',
+            'List',
+            FieldObject::TYPE_NUMERIC,
+            'Contact',
```
gustavs-gutmanis: Lets avoid magic strings. They should all be a single constant `CATEGORY_CONTACT`.

Same goes for the others.

packages/plugin/src/Integrations/CRM/ActiveCampaign/BaseActiveCampaignIntegration.php
```diff
@@ -0,0 +1,572 @@
...
+        if ('Deal' === $category) {
+            return $this->fetchDealFields($client);
+        }
+
+        if ('Account' === $category) {
+            return $this->fetchAccountFields($client);
+        }
+
+        return [];
+    }
```
gustavs-gutmanis: Lets use the new PHP8.0 `match` for such switch cases.
Also please avoid magic strings and use constants instead.

```suggestion
        match ($category) {
            self::CATEGORY_CONTACT => $this->fetchContactFields($client),
            self::CATEGORY_DEAL => $this->fetchDealFields($client),
            self::CATEGORY_ACCOUNT => $this->fetchAccountFields($client),
            default => [],
        }
    }
```

packages/plugin/src/Integrations/CRM/ActiveCampaign/Versions/ActiveCampaignV3.php
```diff
@@ -0,0 +1,387 @@
...
+    )]
+    protected bool $mapContact = false;
+
+    #[Flag(self::FLAG_INSTANCE_ONLY)]
+    #[ValueTransformer(FieldMappingTransformer::class)]
+    #[VisibilityFilter('Boolean(values.mapContact)')]
+    #[Input\Special\Properties\FieldMapping(
+        instructions: 'Select the Freeform fields to be mapped to the applicable ActiveCampaign Contact fields',
+        order: 9,
+        source: 'api/integrations/crm/fields/Contact',
```
gustavs-gutmanis:
```suggestion
        source: 'api/integrations/crm/fields/'.self::CATEGORY_CONTACT,
```

packages/plugin/src/Integrations/CRM/ActiveCampaign/Versions/ActiveCampaignV3.php
```diff
@@ -0,0 +1,387 @@
...
+            $mapping = $this->processMapping($form, $this->contactMapping, 'Contact');
+            if (!$mapping) {
+                return;
+            }
+
+            foreach ($mapping as $key => $value) {
+                if (is_numeric($key)) {
+                    if (\is_array($value)) {
+                        $value = '||'.implode('||', $value).'||';
+                    }
```
gustavs-gutmanis, line 149: Does this apply only for custom fields?
Or would the fields that have non-numeric keys have this same procedure?

There's a `APIIntegrationInterface::EVENT_PROCESS_VALUE` event which lets you modify the value for any given combination of integration fields and freeform fields and value types.

Check out [`Bundles/Integrations/CRM/Salesforce/SalesforceArrayValueProcessor.php`](https://github.com/solspace/craft-freeform/blob/v5/packages/plugin/src/Bundles/Integrations/CRM/Salesforce/SalesforceArrayValueProcessor.php#L36)

packages/plugin/src/Integrations/CRM/ActiveCampaign/Versions/ActiveCampaignV3.php
```diff
@@ -0,0 +1,387 @@
...
+                            }
+                        }
+                    } catch (\Exception $exception) {
+                        $this->processException($exception, self::LOG_CATEGORY);
+                    }
+                } else {
+                    $this->processException($exception, self::LOG_CATEGORY);
+                }
+            }
+        }
```
gustavs-gutmanis: For the sake of better readability and cognitive complexity lets not wrap the whole code inside an `if` condition, but rather fail out if it doesn't match, and continue afterwards if it does.
```suggestion
        if (!$this->account) {
            return;
        }
        
        try {
            $response = $client->post(
                $this->getEndpoint('/accounts'),
                [
                    'json' => [
                        'account' => $this->account,
                    ],
                ],
            );

            $json = json_decode((string) $response->getBody(), false);

            if (isset($json->account)) {
                $this->accountId = $json->account->id;
            }
        } catch (\Exception $exception) {
            if (422 === $exception->getCode()) {
                try {
                    $response = $client->get($this->getEndpoint('/accounts'));

                    $json = json_decode($response->getBody(), false);

                    foreach ($json->accounts as $account) {
                        if (!empty($this->account['name']) && strtolower($account->name) === strtolower($this->account['name'])) {
                            $this->accountId = $account->id;

                            break;
                        }
                    }
                } catch (\Exception $exception) {
                    $this->processException($exception, self::LOG_CATEGORY);
                }
            } else {
                $this->processException($exception, self::LOG_CATEGORY);
            }
        }
    }
```

packages/plugin/src/Integrations/CRM/ActiveCampaign/Versions/ActiveCampaignV3.php
```diff
@@ -0,0 +1,387 @@
...
+                $this->deal['owner'] = (string) $ownerId;
+            }
+
+            if ($this->contactId) {
+                $this->deal['contact'] = (string) $this->contactId;
+            }
+
+            $this->deal['title'] = 'Deal';
+            $this->deal['currency'] = 'usd';
+            $this->deal['value'] = 0;
```
gustavs-gutmanis: I think these would better be set in the `::setProps` since they're set once and are always the same

packages/plugin/src/Integrations/CRM/HubSpot/Versions/HubSpotV1.php
```diff
@@ -10,451 +10,482 @@
...
+    )]
+    protected bool $mapContacts = false;
+
+    #[Flag(self::FLAG_INSTANCE_ONLY)]
+    #[ValueTransformer(FieldMappingTransformer::class)]
+    #[VisibilityFilter('Boolean(values.mapContacts)')]
+    #[Input\Special\Properties\FieldMapping(
+        instructions: 'Select the Freeform fields to be mapped to the applicable HubSpot Contacts fields',
+        order: 8,
+        source: 'api/integrations/crm/fields/Contact',
```
gustavs-gutmanis:
```suggestion
        source: 'api/integrations/crm/fields/'.self::CATEGORY_CONTACT,
```

packages/plugin/src/Integrations/CRM/HubSpot/Versions/HubSpotV1.php
```diff
@@ -10,451 +10,482 @@
...
-
-    #[Flag(self::FLAG_ENCRYPTED)]
-    #[Flag(self::FLAG_GLOBAL_PROPERTY)]
-    #[Validators\Required]
-    #[Input\Text(
-        label: 'API Key',
-        instructions: 'Enter your HubSpot API key here.',
-    )]
-    protected string $apiKey = '';
+    // ==========================================
```
gustavs-gutmanis, line 38:
```php
   protected const CATEGORY_CONTACT = 'contacts';
   protected const CATEGORY_DEAL = 'deals';
   protected const CATEGORY_COMPANY = 'companies';
```

packages/plugin/src/Integrations/CRM/HubSpot/Versions/HubSpotV1.php
```diff
@@ -10,451 +10,482 @@
...
+    public function fetchFields(string $category, Client $client): array
     {
-        return $this->appendContactData;
+        $record = 'deals';
+
+        if ('Contact' === $category) {
+            $record = 'contacts';
+        } elseif ('Company' === $category) {
+            $record = 'companies';
+        }
```
gustavs-gutmanis: Please no magic strings.

Also there's no need for this switch, just specify `contacts` as the category, rather than `Contact`.

packages/plugin/src/Integrations/CRM/HubSpot/Versions/HubSpotV1.php
```diff
@@ -10,451 +10,482 @@
...
-                    break;
+            foreach ($this->contactMapping as $fieldMapping) {
+                if ('' === $fieldMapping->getValue()) {
+                    continue;
+                }
 
-                case 'company':
-                    $companyProps[] = ['value' => $value, 'name' => $propName];
+                $formFields[$fieldMapping->getSource()] = $form->getLayout()->getField($fieldMapping->getValue());
+            }
```
gustavs-gutmanis: The whole reason for the new field mappings is to avoid having to write such custom code here.

You can listen to the `APIIntegrationInterface::EVENT_PROCESS_VALUE` and make any changes you need there, without having ever to touch this process.

Then you can only extract the values via `$this->processMapping` and work with them.

packages/plugin/src/Integrations/CRM/Insightly/Versions/InsightlyV31.php
```diff
@@ -0,0 +1,86 @@
...
+        return true;
+    }
+
+    private function processLeads(Form $form, Client $client): void
+    {
+        if (!$this->mapLeads) {
+            return;
+        }
+
+        $mapping = $this->processMapping($form, $this->leadMapping, 'Lead');
```
gustavs-gutmanis: No magic strings pls

packages/plugin/src/Integrations/CRM/Pardot/Versions/PardotV4.php
```diff
@@ -0,0 +1,129 @@
...
+    )]
+    protected bool $mapProspect = false;
+
+    #[Flag(self::FLAG_INSTANCE_ONLY)]
+    #[ValueTransformer(FieldMappingTransformer::class)]
+    #[VisibilityFilter('Boolean(values.mapProspect)')]
+    #[Input\Special\Properties\FieldMapping(
+        instructions: 'Select the Freeform fields to be mapped to the applicable Pardot Prospect fields',
+        order: 6,
+        source: 'api/integrations/crm/fields/Prospect',
```
gustavs-gutmanis: No magic pls 🪄

packages/plugin/src/Integrations/CRM/Zoho/BaseZohoIntegration.php
```diff
@@ -206,18 +206,16 @@ protected function getDomain(): string
...
     {
         $data = $response['data'][0];
+
         if ('error' === $data['status']) {
             $this->getLogger()->error(
                 $data['message'],
-                ['exception' => $data],
+                [
+                    'exception' => $data,
+                ],
```
gustavs-gutmanis: If there's just one element in the array, it looks a bit less crowded in a single line.

gustavs-gutmanis: Unfortunately I cannot enforce this rule with the PHP-CS-Fixer

packages/plugin/src/Library/Integrations/BaseIntegration.php
```diff
@@ -100,12 +102,30 @@ protected function getLogger(?string $category = null): LoggerInterface
...
+                    $usefulErrorMessage = (string) $response->getBody();
+                }
+            }
+        }
+
+        $this->getLogger($category)->error(
+            $usefulErrorMessage,
+            [
+                'exception' => $exception->getMessage(),
+            ],
```
gustavs-gutmanis: `err` and `error` checks seem super specific to a given integration.
Almost no two integrations have the same error reporting in their responses, so I would have the `BaseIntegration::processException` only deal with the most basic of processing - the exception message.

And then if an Integration (such as Salesforce) has specific error reporting, we can override this method in that class.

packages/plugin/src/Library/Integrations/Types/CRM/CRMIntegration.php
```diff
@@ -26,6 +29,77 @@ public function getType(): string
...
+                    $value = [$value];
+                }
+
+                return $value;
+
+            case FieldObject::TYPE_STRING:
+            default:
+                return (string) $value;
+        }
+    }
```
gustavs-gutmanis: This method is not needed. We have this processing going on in the [`CrmBundle`](https://github.com/solspace/craft-freeform/blob/v5/packages/plugin/src/Bundles/Integrations/CRM/CrmBundle.php#L85).

Actually in my PR #767 it is moved into `IntegrationsBundle` and is used by both CRM and MailingList integrations.

packages/plugin/src/controllers/api/integrations/CrmController.php
```diff
@@ -10,6 +11,15 @@
 
 class CrmController extends BaseApiController
 {
+    public function __construct(
+        $id,
+        $module,
+        private IntegrationClientProvider $clientProvider,
+        $config = [],
```
gustavs-gutmanis: This might cause issues, the `$config` argument is supposed to be the third one.

packages/plugin/src/Services/Integrations/IntegrationsService.php
```diff
@@ -103,7 +105,8 @@ public function getByHandle(string $handle): ?IntegrationModel
     public function save(IntegrationModel $model, IntegrationInterface $integration, bool $triggerEvents = false): bool
     {
         try {
-            $integration->onBeforeSave();
+            $client = $this->clientProvider->getAuthorizedClient($integration);
+            $integration->onBeforeSave($client);
```
gustavs-gutmanis: I'm not entirely sure about this approach.
For OAuth integrations the `$client` will definitely not work when saving.

Since this is only for one integration, I think it doesn't even need to have those stage, etc values before its pushing them. You might fetch them only during the request. 
Or build out the client manually on before save, since they're using API tokens.

packages/plugin/src/Integrations/CRM/ActiveCampaign/Versions/ActiveCampaignV3.php
seandelaney: @gustavs-gutmanis I'm not entirely sure. I carried this over from the existing integration and I didn't want to touch it.

packages/plugin/src/Library/Integrations/Types/CRM/CRMIntegration.php
seandelaney: @gustavs-gutmanis 

Okay. How does this function call work then:

https://github.com/solspace/craft-freeform/blob/0eec3eeffad11b7e42c3284cf817f4add4ea856c/packages/plugin/src/Integrations/CRM/Pardot/BasePardotIntegration.php#L127

seandelaney: I don't understand this yet and can't see where or how its used.

seandelaney: I see you updated the EmailMarketingBundle in your PR, completely removing convertCustomFieldValue function but if I remove these calls in Pardot and Infusionsoft, both will cause issues until your PR is merged. I guess that's okay...

packages/plugin/src/Integrations/CRM/Pardot/Versions/PardotV4.php
seandelaney: But I can do magic.

packages/plugin/src/Integrations/CRM/Insightly/Versions/InsightlyV31.php
seandelaney: Pleeeeease?

gustavs-gutmanis (approved):
Nice work, lets get this merged.

PR #769, opened 2023-08-14
[SFT-512] - Update Freshdesk CRM

seandelaney:


packages/plugin/src/Integrations/CRM/Freshdesk/Versions/FreshdeskV2.php
```diff
@@ -0,0 +1,174 @@
...
+        instructions: 'Select the Freeform fields to be mapped to the applicable Freshdesk Ticket fields',
+        order: 8,
+        source: 'api/integrations/crm/fields/'.self::CATEGORY_TICKET,
+        parameterFields: ['id' => 'id'],
+    )]
+    protected ?FieldMapping $ticketMapping = null;
+
+    private array $values = [];
+
+    private array $customValues = [];
```
gustavs-gutmanis: It seems like these two are only used within the `::processTickets()` method and thus not really necessary outside of its scope.

seandelaney: Good catch

gustavs-gutmanis (approved):
Looking good 👌🏻

PR #772, opened 2023-08-15
Email Marketing: [SFT-517] ActiveCampaign and [SFT-519] Constant Contact

seandelaney:
Also...

- Fixed a routing issue with non OAuth mailing list integrations.
- Added refresh button to `Dynamic Select` field type. 
  - Field would attempt to gab mailing lists load upon page but then never show any new lists. Hard refreshing the browser felt dirty.
- Changed `EmailMarketing` to `MailingLists` for consistency.


packages/plugin/src/Integrations/CRM/ActiveCampaign/BaseActiveCampaignIntegration.php
```diff
@@ -283,8 +286,9 @@ private function fetchContactFields(Client $client, string $category): array
 
         foreach ($json->fields as $field) {
             $type = match ($field->fieldType) {
-                'dropdown', 'multiselect', 'checkbox' => FieldObject::TYPE_ARRAY,
-                'date' => FieldObject::TYPE_DATETIME,
+                'text', 'textarea', 'hidden', 'radio' => FieldObject::TYPE_STRING,
```
gustavs-gutmanis: These already fall into the `default` category and would get `FieldObject::TYPE_STRING` assigned to them, no?

packages/plugin/src/Integrations/MailingLists/Mailchimp/BaseMailchimpIntegration.php
```diff
@@ -23,19 +23,19 @@
...
 
-abstract class BaseMailChimpIntegration extends MailingListIntegration implements OAuth2ConnectorInterface, MailChimpIntegrationInterface
+abstract class BaseMailchimpIntegration extends MailingListIntegration implements OAuth2ConnectorInterface, MailchimpIntegrationInterface
 {
     use OAuth2Trait;
 
     protected const LOG_CATEGORY = 'Mailchimp';
 
-    protected const CATEGORY_MEMBERS = 'Members';
+    protected const CATEGORY_CONTACT = 'Contact';
```
gustavs-gutmanis, line 32: What's up with the renames here?
My idea behind the plurals is that you map form "submissions" to "members", thus plural. But maybe that hasn't been consistent across the project.

And `MEMBERS` because Mailchimp in particular refers to them as audience members.

Anyway, would love to hear your thoughts on this, just wondering.

gustavs-gutmanis (approved):

packages/plugin/src/Integrations/CRM/ActiveCampaign/BaseActiveCampaignIntegration.php
seandelaney: Great catch!

packages/plugin/src/Integrations/MailingLists/Mailchimp/BaseMailchimpIntegration.php
seandelaney: Yep, wasn't consistent. Some used plurals, some didn't.

Then Kelsey updating the language to use contacts instead of members, as Mailchimp docs and dashboard refers to contacts and not members, so also changed for consistency.

PR #773, opened 2023-08-16
[SFT-501]: Add Option field type populators back (Elements & Predefined)

gustavs-gutmanis:
* adding element option value populators to all option type fields.
* adding `Entry`, `Category`, `User` and `Tag` elements for now.


seandelaney (approved):
Nice, all works

PR #774, opened 2023-08-17
[SFT-518] - Update Campaign Monitor Email Marketing and [SFT-520] - Campaign plugin Email Marketing

seandelaney:


kjmartens:
Good enough for now... will log some bugs separately 😄

PR #776, opened 2023-08-21
[SFT-551]: Restyle custom dropdown

gustavs-gutmanis:
* updating dropdown styles
* adding nested group fields
* adding group icons in dropdowns


kjmartens (approved):
Enchanting!

seandelaney (approved):
Bit of an overflow issue so not sure if we can do anything about that yet
image

Otherwise, all working nicely!

PR #778, opened 2023-08-22
[SFT-524] - Add red asterisk * to required fields inside of form builder Layout

seandelaney:
Added a more strict required element since `.required` class was being controlled/inherited badly by multiple freeform demo styles and causing an issue. This way, no conflicts or loss of required if freeform demo removed/updated.


packages/client/src/app/components/form-controls/control.styles.ts
```diff
@@ -7,10 +7,23 @@ type LabelProps = {
...
+
+  .is-required {
+    display: flex;
+    margin-top: 0;
+    margin-left: 2px;
+    flex-direction: row;
+    color: ${colors.error};
+    align-items: flex-start;
+    justify-content: flex-start;
+  }
```
gustavs-gutmanis: Maybe I could suggest using the `:after` pseudo element for this, rather than writing so much css?

```ts
export const Label = styled.label<LabelProps>`
  display: block;

  color: ${colors.gray550};
  font-weight: ${({ regular }) => (regular ? 'normal' : 'bold')};
  
  &.is-required {
    &:after {
      content: '*';
      padding-left: 2px;

      color: ${colors.error};
    }
  }
```

packages/client/src/app/components/form-controls/label.tsx
```diff
@@ -23,7 +23,7 @@ const FormLabel: React.FC<Props> = ({ label, handle, required, title }) => {
           {label}
         </Tooltip>
       )}
-      {required && <span className="required" />}
+      {required && <span className="is-required">&#42;</span>}
```
gustavs-gutmanis: instead of this, you would add the class directly to `<Label />`

```tsx
<Label
  className={classes(required && 'is-required')}
  htmlFor={handle}
>
```

packages/client/src/app/components/form-controls/control.styles.ts
seandelaney: @gustavs-gutmanis I had tried this approach but inheritance from something else is breaking the layout/alignment of the asterisk via ::after. I will give it another go though.

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-layout/field/cell/cell.styles.ts
```diff
@@ -4,8 +4,18 @@ import styled from 'styled-components';
...
 
-  color: ${colors.gray550};
   font-weight: bold;
+  color: ${colors.gray550};
+
+  &.is-required {
+    &:after {
+      content: '*';
+      padding-left: 2px;
+      position: absolute;
```
gustavs-gutmanis, line 14: How does `position: absoulte;` work here?
What holds the relative position?

seandelaney: There is `LoadingText` element, which uses flex properties, all effecting the alignment. This is why I went down the strict route in the description notes. But position: absolute seems to fix the mis-alignment.

gustavs-gutmanis (approved):

PR #780, opened 2023-08-22
[SFT-535] - Rename settings properties to no longer include general and behavior.

seandelaney:
```
form.settings.ajax,
form.settings.general.handle,
form.settings.behavior.limitSubmissions,
form.settings.general
```


kjmartens:
Works good for me 🙂

PR #782, opened 2023-08-23
[SFT-495]: Bring in reCAPTCHA/hCaptcha to Integrations builder tab

gustavs-gutmanis:
This also takes care of SFT-507

* adding recaptcha
* adding hcaptcha
* adding `captchas` integration category
* introducing feature bundle search to integrations directories, so that each integration can have its own bundles nested into their directories
* removed all traces of old recaptcha implementations
* refactored code logic for captchas into their own JS files which employ freeform form events


kjmartens (approved):
Looks delicious! It seems to work as expected. I adjusted a few things like setting order/default/wording.

packages/plugin/src/Form/Form.php
```diff
@@ -742,24 +736,6 @@ public function jsonSerialize(): array
...
-                'enabled' => true,
-                'handle' => 'reCaptcha',
-                'name' => $isHCaptcha ? 'h-recaptcha-response' : 'g-recaptcha-response',
-            ];
-        } else {
-            $object['reCaptcha'] = [
-                'enabled' => false,
-            ];
-        }
-
```
seandelaney, line 762: Lets remember to check this again when we update GraphQL, as this was used in the Form queries...I see no comment to refactor

PR #784, opened 2023-08-24
[SFT-536] - If a Date field is left empty, it breaks the submit validation (when not required)

seandelaney:
Added logic to skip if no value found. 

Note, if the field was required and submitted empty, the required validation logic will have already pushed back an error message. So this new logic will only trigger if not required and no value.


gustavs-gutmanis (approved):
Lookin' good!

PR #786, opened 2023-08-24
[SFT-530]: Switch form, integration, and field name generators to just be simple, e.g. Form 1, Form 2, Textarea 1, Textarea 2, etc

gustavs-gutmanis:
* generating more generic names for new fields and notifications


kjmartens (approved):
Looks good

PR #789, opened 2023-08-24
[SFT-497]: Bring in Webhooks to Integrations builder tab

gustavs-gutmanis:
* adding webhooks as integrations


kjmartens (changes requested):
It looks good but the builder "Message" textarea for Slack is just 2 rows (10+ would be nice), and the "Message" textarea for setting defaults in the Settings -> Webhooks area is just a single line input instead of textarea.

seandelaney (approved):
Tested and worked nice. I got the message in Slack too.

PR #794, opened 2023-08-25
[SFT-538] - Phone fields don't do JS validation on front end

seandelaney:
- Updated logic to skip validation if no value was submitted and not required
- Applied correct attributes to input so phone JS validation worked


kjmartens (approved):
Looks good to me

PR #796, opened 2023-08-25
[SFT-506]: Add back HTML and Rich Text field types

gustavs-gutmanis:
* adding `HTML` field editor
* adding `Rich Text` field editor
* displaying both fields properly in form renders


kjmartens (changes requested):
Looks good but the following should happen:

- Remove Instructions setting
- Remove Required checkbox
- Adjust the height of the editor popout for smaller screens

seandelaney (approved):
Works nicely. Tested with a full webpage copy and paste HTML dump

PR #797, opened 2023-08-28
[SFT-509]: Freeform 5 seems to get confused when "Custom Values" checkbox ISN'T used on Option fields.

gustavs-gutmanis:
* fixing custom option values not being populated when `use custom values` is turned off
* fixing positioning of the editor pop-ups


seandelaney (approved):
Working correctly

PR #800, opened 2023-08-28
[SFT-543]: The form builder Settings tab does not load the section segment in URL and doesn't always load the form if you refresh the browser.

gustavs-gutmanis:
* fixing form settings navigation
* url changes when switching sections
* window reloaded url opens correct section
* opening blank sections triggers the first section
* adding loaders for loading state


seandelaney (approved):
It's the little things that make life easier.

PR #801, opened 2023-08-28
[SFT-544]: Conditional Rules (Rules tab) need to have ability to delete the condition. Currently there is no way

gustavs-gutmanis:
* adding remove buttons for rules


seandelaney (approved):
Beautiful

PR #802, opened 2023-08-28
[SFT-546]: For "Rules", any implicated field should have a light gray corner "always", but then have the corner be green inside the context of/viewing a field's rules

gustavs-gutmanis:
All fields show a gray notch if they're used in any conditions


kjmartens (changes requested):
Looks good, but it seems that rules on PAGES do not correctly add corner color to fields that affect it. Happens fine for all other stuff.

Screenshot 2023-08-28 at 10 00 12 AM

seandelaney (commented):
I see the same as K. No color corner on Pages rules. Everything else is working fine.

PR #803, opened 2023-08-28
[SFT-559]: improving opinion scales and legends

gustavs-gutmanis:
* legends are now `Legend` objects which convert to string values
* scales are now `Scale` objects which have `::getLabel()` and `::getValue()` methods as well as conversion to label when output as string


packages/plugin/src/Attributes/Property/Implementations/OpinionScale/LegendsTransformer.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public function reverseTransform($value): array
+    {
+        $data = [];
+        if (!\is_array($value)) {
+            return $data;
+        }
+
+        // @var Legend $leged
```
seandelaney: typo

seandelaney (approved):

kjmartens (changes requested):
Looks good, but is it possible to have `{{ scale }}` become the equivalent of `{{ scale.label ?: scale.value }}` so that people not using labels are still covered?

PR #804, opened 2023-08-28
[SFT-561] - The File & File DnD field types are missing the Asset location and Upload Location Subfolder settings inside the form builder

seandelaney:
- Added field configuration for assets source id.
- Added field configuration for default location path.
- Fixed an issue where hasFields would return an object with an inner `fields` property resulting in boolean logic to return incorrectly. Resulting in the form missing `enctype` attribute.
- Fixed typo in DnD field attributes preventing users from dragging and dropping fields.
- Fixed an issue where checkboxes control type was throwing an error due to missing `onChange` even handler.
- Fixed an issue where the default asset source id was never being set upon loading.
- Fixed an issue with missing class name, resulting in drog and drag area was not defined.


kjmartens (changes requested):
Seems mostly good, but I have 2 file upload fields (one File and one File DnD) in my form. Whichever field occurs first will get the files uploaded correctly, and the second field will get nothing uploaded at all. I also rearranged the order of the file fields in the layout and same behavior -- first field works, second does not.

seandelaney:
@kjmartens I've updated the PR again, improving some bits. 

I can now upload files using both File Upload field, File DnD Field and a combo of both fields. 

However, when using a combo of both files, the files added to the DnD are uploaded twice and I cannot seem to find where this is happening.

I'm hoping once @gustavs-gutmanis reviews/tests, he can spot it.

packages/plugin/src/Bundles/Fields/Validation/FileUploadValidation.php
```diff
@@ -58,7 +57,7 @@ public function __construct(
     public function uploadFiles(ValidationEvent $event): void
     {
         $form = $event->getForm();
-        $uploadFields = $form->getLayout()->getFields(FileUploadInterface::class);
+        $uploadFields = $form->getLayout()->getFields(FileUploadField::class);
```
gustavs-gutmanis, line 60: Wondering why not use the interface here?

packages/plugin/src/Fields/Implementations/ValueGenerators/AssetSourceGenerator.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    public function generateValue(Property $property, string $class, ?object $referenceObject): ?int
+    {
+        $volumes = \Craft::$app->getVolumes()->getViewableVolumes();
+
+        if (\count($volumes)) {
+            return $volumes[0]->id;
+        }
+
+        return null;
```
gustavs-gutmanis, line 28: Probably can be written as something like this..

```suggestion
        return $volumes[0]->id ?? null;
```

gustavs-gutmanis (approved):
Looking tight

gustavs-gutmanis:
> @kjmartens I've updated the PR again, improving some bits.
> 
> I can now upload files using both File Upload field, File DnD Field and a combo of both fields.
> 
> However, when using a combo of both files, the files added to the DnD are uploaded twice and I cannot seem to find where this is happening.
> 
> I'm hoping once @gustavs-gutmanis reviews/tests, he can spot it.

I'll check this out in a minute

packages/plugin/src/Bundles/Fields/Validation/FileUploadValidation.php
seandelaney: @gustavs-gutmanis 

See this comment: https://github.com/solspace/craft-freeform/pull/804#issuecomment-1699694630

I changed to `FileUploadField` in an attempt to exclude `FileDragAndDropField` but it doesn't make any difference as `FileDragAndDropField` extends `FileUploadField`.

So it can be changed back to `FileUploadInterface`.

gustavs-gutmanis:
I could not get the duplicate upload to happen. I had both the file upload and D&D upload fields and both had files uploaded without issues.

seandelaney:
I guess if @kjmartens cannot replicate, then we can merge

kjmartens (approved):

PR #805, opened 2023-08-29
[SFT-557] - Fixed a bug where Logged in Users only limitation was still allowing guests. Plus some cleanup

seandelaney:


gustavs-gutmanis (approved):
Looks good.
Probably some of the methods can be eventually combined so as not to duplicate the code, but that can be done later.

PR #808, opened 2023-08-29
[SFT-583]: Forms ALWAYS seem to submit as AJAX no matter if the AJAX setting is enabled or not

gustavs-gutmanis:
* removing stripe handler, since we don't have stripe right now
* thus fixing ajax always being on


kjmartens (changes requested):
reCAPTCHA
- Original issue appears to be fixed.
- New issue on multipage forms where (v2 Invis):
    - On second page I get error: `2recaptcha__en.js:179 Uncaught Error: No reCAPTCHA clients exist. at Array.<anonymous> (recaptcha__en.js:179:417) at recaptcha__en.js:407:241`
    - On second/additional pages I notice the Previous and Submit buttons are labeled as `undefined`

kjmartens (approved):

seandelaney (commented):
I also approve. I had started reviewing and it was merged mid my review. lol

PR #812, opened 2023-08-30
[SFT-569] - Can we update any references to 'Select' in Table fields …

seandelaney:
…- builder and CODE/params, etc - to 'Dropdown' for consistency


kjmartens (approved):
Nice!

PR #813, opened 2023-08-30
[SFT-598] - Adjust Attribute editor

seandelaney:


gustavs-gutmanis (approved):
Maybe it's better to do the ones which should show their container, and do `...` for the rest?

PR #814, opened 2023-08-31
[SFT-570] - Field button names (left column in Layout) in form builder are being clipped off 1-2px at bottom

seandelaney:


kjmartens (approved):
Tested, looks great

PR #816, opened 2023-08-31
[SFT-571] - Update all single checkbox settings inside property editor to use Toggle instead

seandelaney:


kjmartens (approved):
Looks good, minus the part where the text is darker, but I'll make a different task for that

PR #817, opened 2023-08-31
[SFT-562] - Notifications tab inside builder will show loading state forever when there are no notifications configured

seandelaney:


kjmartens (approved):
Looks good enough for now. 🙂

PR #818, opened 2023-08-31
[SFT-573] - Table field preview in builder should look similar to Craft's Table field

seandelaney:


kjmartens (approved):
Looks wonderful!! 👌

PR #821, opened 2023-09-01
[SFT-609] - Change Label Column Type in CRM Fields

seandelaney:
https://solspace.atlassian.net/browse/SFT-609


gustavs-gutmanis (approved):
I guess this makes sense

PR #823, opened 2023-09-04
[SFT-564]: In form builder, all fields are missing "default value" setting

gustavs-gutmanis:
* adding `Default Value` fields to all eligible fields in builder
* using `defaultValue` in the respective places


kjmartens (changes requested):
1. Toggling default values on Checkboxes fields will toggle ALL options on/off at once.
2. Setting a default option for Radio fields works but doesn't show visually on that field in the form builder.
3. Single checkbox fields always show as checked by default when rendered, even if the setting is not applied for it.

seandelaney (commented):
image

Toggling options seems to use label value. I know there shouldn't be multiple labels the same but should the checked state not use index position / key ?

gustavs-gutmanis:
It uses the Option `value`, which in this particular instance is the same as `label` since they're not using custom values.

The `defaultValue` will be an array of `["Option 4", "Option 4"]` which will result in `["Option 4"]` since there are no duplicate values in the value.

If we were to use indices instead of values, adding a value in the middle of a list would cause all of the values to shift. While there are workarounds - the values would be unreliable.

Imagine a scenario where someone submits checkboxes of indices `[1, 4, 6]`. Now after a while someone changes the list a little bit, to have more or less values, and reorders them.
Now all of the previously submitted values are no longer valid.

If we store the actual values, but use indexes in the builder, then we have an unnecessary data misalignment, and need to write extra logic everywhere to go back and forth.

PR #824, opened 2023-09-05
[SFT-601]: When submitting a form without AJAX, the 'row', 'success', 'errors' attributes seem to add more and more duplicates for each submit of the form (and errors present)

gustavs-gutmanis:
* skipping attribute merge on posted forms, since they have the attributes applied already


kjmartens (changes requested):
1. Confirm fields seem to be omitted from the `Attributes` handling. I have nothing discriminating against them and they show up without any of the classes they should have. Screenshot 2023-09-05 at 9 28 24 AM
2. Confirm fields do not show up inside the CP Submissions detail view - probably still omitted from old approach.
3. Custom formatting templates include the complete file path, instead of relative to the Formatting Templates directory set in the FF settings. 
Screenshot 2023-09-05 at 9 44 56 AM

kjmartens (approved):

PR #825, opened 2023-09-05
[SFT-580] - ActiveCampaign CRM integration does not load anything for Contacts mapping

seandelaney:
Custom fields with numeric key, e.g `1` where being pulled from API and saved but then getting deleted again as `in_array` strict check was failing due to numeric key value.

Also contains fixes for https://solspace.atlassian.net/browse/SFT-578 as related


packages/plugin/src/Integrations/CRM/ActiveCampaign/BaseActiveCampaignIntegration.php
```diff
@@ -295,20 +289,22 @@ private function fetchDealFields(Client $client, string $category): array
...
+        if (isset($json->dealCustomFieldMeta)) {
+            foreach ($json->dealCustomFieldMeta as $field) {
+                $type = match ($field->type) {
+                    'dropdown', 'multiselect', 'checkbox' => FieldObject::TYPE_ARRAY,
+                    'date' => FieldObject::TYPE_DATETIME,
+                    default => FieldObject::TYPE_STRING,
+                };
+
+                $fieldList[] = new FieldObject(
+                    'deal__'.$field->id,
```
gustavs-gutmanis: is this what you mentioned in the description?

gustavs-gutmanis: would it maybe make more sense to get integer fields working rather than doing these checks everywhere?

seandelaney: @gustavs-gutmanis I can attempt to get integer fields working but this was the approach used in v4, which I had took out in refactor.

gustavs-gutmanis: Yes, we should not be prefixing anything if at all possible, that just puts way too much burden on us and should not be the way to go about this

seandelaney: @gustavs-gutmanis Integer field ids now working as expected.

gustavs-gutmanis (approved):
Great stuff

PR #826, opened 2023-09-06
[SFT-556]: The Stop Submissions After Date setting seems to automatically insert today's date in it and has no option to delete the value

gustavs-gutmanis:
* improved datepicker components
* fixing issues with "stop submissions after date" setting


kjmartens (approved):
Enchanting

PR #827, opened 2023-09-06
[SFT-612]: When using `"successBehavior":"reload"` AND no AJAX, it doesn't seem to do that and uses the `return URL` instead

gustavs-gutmanis:
* improving success behavior handling


kjmartens (approved):
AMAZE

PR #833, opened 2023-09-08
[SFT-615]: Attempting to edit an existing submission on front end (demo templates) triggers error

gustavs-gutmanis:
* fixing element editing via Freeform forms


kjmartens (dismissed):
All good, but this was not regarding editing other elements (if that's what you were fixing). In my case, I was getting a `Call to undefined method Solspace\Freeform\Form\Layout\FormLayout::getStorableFields()` error when attempting to load the EDIT page in the front end/demo templates, and it's still happening for me.

gustavs-gutmanis:
Woops, ok now it's fixed.

PR #835, opened 2023-09-11
[SFT-635]: Creating a new Export Profile doesn't work, receive "Call to undefined method Solspace\Freeform\Form\Types\Regular::getForm()" error.

gustavs-gutmanis:
* fixing export profiles


kjmartens (approved):
Looks good

PR #836, opened 2023-09-11
[SFT-625]: Integration field mapping (and some Settings areas?) should use custom dropdown field setting

gustavs-gutmanis:
* fixing spacing issues in integrations and user select options
* introducing dropdown to many places


kjmartens (changes requested):
Looks excellent! But I guess we need to resolve the positioning issue 😅 

Screenshot 2023-09-11 at 9 47 38 AM

PR #841, opened 2023-09-13
[SFT-619] - Updated GraphQL to support Freeform v5

seandelaney:


packages/plugin/src/Bundles/GraphQL/Interfaces/FieldInterface.php
```diff
@@ -46,10 +47,14 @@ public static function getFieldDefinitions(): array
...
             'hash' => [
                 'name' => 'hash',
                 'type' => Type::string(),
                 'description' => "Field's hash",
+                'resolve' => function () {
+                    return 'This property is deprecated';
+                },
             ],
             'handle' => [
                 'name' => 'handle',
```
gustavs-gutmanis, line 50: Is this supposed to be `handle`?

packages/plugin/src/Bundles/GraphQL/Interfaces/FieldInterface.php
```diff
@@ -75,30 +80,65 @@ public static function getFieldDefinitions(): array
...
+                    return self::transform($source->getAttributes()->getInput());
+                },
             ],
             'labelAttributes' => [
                 'name' => 'labelAttributes',
                 'type' => Type::listOf(AttributeInterface::getType()),
                 'description' => "Field's label attributes",
+                'resolve' => function ($source) {
+                    return self::transform($source->getAttributes()->getLabel());
+                },
```
gustavs-gutmanis: These attributes are how FF4 worked.
It would be nice if we had `attributes` which resolves into an object with `input, container, error, etc` instead of having them in the root of the object.

packages/plugin/src/Bundles/GraphQL/Interfaces/FormInterface.php
```diff
@@ -111,61 +149,112 @@ public static function getFieldDefinitions(): array
...
             ],
             'gtmEventName' => [
                 'name' => 'gtmEventName',
                 'type' => Type::string(),
                 'description' => 'The name of the Event that will be added to Google Tag Manager\'s data layer ',
+                'resolve' => function ($source) {
+                    // FIXME
+                    return null;
+                },
             ],
```
gustavs-gutmanis: These can be safely removed

packages/plugin/src/Fields/AbstractField.php
```diff
@@ -43,6 +43,18 @@
...
+        instructions: "How you'll refer to this field in templates",
+        order: 2,
+        placeholder: 'myField',
+    )]
+    #[Middleware('handle')]
+    #[Flag('code')]
+    #[Validators\Required]
+    #[Validators\Handle]
+    #[Validators\Length(100)]
+    public string $handle = '';
```
gustavs-gutmanis: Label should precede handle.
And why is handle made public here?

There's a `getHandle()` method for getting the handle.

packages/plugin/src/Form/Settings/Implementations/GeneralSettings.php
```diff
@@ -106,6 +106,22 @@ class GeneralSettings extends SettingsNamespace
...
+    )]
+    public string $postForwardingUrl = '';
+
+    #[Section(self::SECTION_GENERAL)]
+    #[Input\Text(
+        label: 'POST Forwarding Error Trigger Phrase',
+        instructions: 'Provide a keyword or phrase Freeform should check for in the output of the external POST URL to know if and when there’s an error to log, e.g. ‘error’ or ‘an error occurred’.',
+        order: 9,
+    )]
+    public string $postForwardingErrorTriggerPhrase = '';
```
gustavs-gutmanis: These will not be here anymore. They'll be an integration instead.

gustavs-gutmanis (commented):
Adding some comments

packages/plugin/src/Fields/AbstractField.php
seandelaney: GraphQL was throwing an error that it couldn't access `handle`

packages/plugin/src/Bundles/GraphQL/Interfaces/FieldInterface.php
seandelaney: @gustavs-gutmanis I'm no sure what you mean?

gustavs-gutmanis: It seems like I mixed things up, because of code collapsing, never mind.

packages/plugin/src/Fields/AbstractField.php
gustavs-gutmanis: Can you write a resolver to use the `::getHandle()` method?
We should not make this property public just because of GQL.

seandelaney: I tried this already.

Here is the error 

```
message: "Cannot access protected property Solspace\\Freeform\\Fields\\Implementations\\TextField::$handle"
```

I will switch back to protected and see if I can find a way around this. GraphQL is not happy with protected for some reason.

seandelaney: Added fix

gustavs-gutmanis (approved):
Beautiful work

PR #844, opened 2023-09-14
[SFT-450]: Improve variables passed to twig render functions for form and fields

gustavs-gutmanis:
* refactored template locations. moved them out of `templates/_defaultFormattingTemplates` into `templates/_templates/formatting` along with success and notification templates.
* refactored test namespace and location
  * this gives automatic discovery of test folders to composer and IDE's
  * moved from `packages/plugin/tests` into `packages/plugin/src/Tests`
  * switched namespace from `Solspace\Tests\Freeform\Unit` to `Solspace\Freeform\Tests`
* refactored the way attributes are passed down to fields and page buttons
  * instead of passing everything in the `attributes`, there's a specific global namespace `fields` and `buttons`
  * you can now override field properties rather than just attributes
  * this processor also handles converting basic data into actual PHP types by using the Transformers that are attached to properties already.


seandelaney (approved):
Looks okay. Cannot find any errors

kjmartens (approved):

PR #845, opened 2023-09-14
Payment Integrations - Stripe - Added support for Phone Field Type in Customer Field Mapping

seandelaney:


gustavs-gutmanis (approved):
Gorgeous!

kjmartens (approved):
Tested, works well.

PR #846, opened 2023-09-15
[SFT-179]: Refactor the way `value` / `valueAsString` is handled so there are more intuitive options to get option label vs option value

gustavs-gutmanis:
* adding `::labels()` and `::labelsAsString()` methods to all option fields.
* adding `::implements()` method to all fields for twig friendly implementation checks
* refactoring `::valueAsString` to not take a parameter


kjmartens (approved):
Seems to work enchantingly

PR #849, opened 2023-09-19
[SFT-596]: When setting Rules logic for a field, if the field used in a condition (from dropdown to choose a field) contains CUSTOM options, the whole builder crashes

gustavs-gutmanis:
* fixing `shadowIndex` issue on production scripts


kjmartens (approved):
Splendid!

PR #851, opened 2023-09-19
[SFT-660]: Email notifications for User Select type do not seem to work at all

gustavs-gutmanis:
* fixing dynamic notifications


kjmartens (changes requested):
The only one I can get to work at all is a Radio field that is populated by Elements (Users). I have other field types and have a variety of options types (Custom, Predefined), and none seem to work.

For a Checkboxes field, I have "Default Recipients" filled out with an email address (in the event no field options were selected), and that one gets triggered every time as if I did not select any option. Option values for this one are Custom and have values like `item-1`, etc.

seandelaney (dismissed):
1 - I have an Opinion scale field in my form and that field is also listed in the Target field dropdown - doesn’t seem to do anything. Values not populated.

2 - Also a single Checkbox field, checked by default with a default value. Field is listed but again, doesn’t seem to do anything. Value not populated.

3 - I have a Radio button field, Source is Elements --> Users.  Option Label set to Full Name but one of my users dont have a full name set. So this user is missing from the list. This field is listed as a target field and populated only users with full names setup.

If I update Radio button field source to Predefined list and select Languages, use native languages, some have blank rows - assuming the native language characters are fucked up a bit. Seems to be Luba-Katanga the problem here.

image (1)

image

Sorry, they are real unique edge cases, well, defo that last one anyways!

But other than the Radio button weirdness with User Full Names and Languages, the only field I cannot get working (maybe failing to understand how it works) is the single checkbox field.

kjmartens (approved):

kjmartens:
> 1 - I have an Opinion scale field in my form and that field is also listed in the Target field dropdown - doesn’t seem to do anything. Values not populated.
> 
> 2 - Also a single Checkbox field, checked by default with a default value. Field is listed but again, doesn’t seem to do anything. Value not populated.
> 
> 3 - I have a Radio button field, Source is Elements --> Users. Option Label set to Full Name but one of my users dont have a full name set. So this user is missing from the list. This field is listed as a target field and populated only users with full names setup.
> 
> If I update Radio button field source to Predefined list and select Languages, use native languages, some have blank rows - assuming the native language characters are fucked up a bit. Seems to be Luba-Katanga the problem here.
> 

@seandelaney I'll log these as a separate issue so we can close out this PR, but thanks for uncovering them 🙂

PR #856, opened 2023-09-28
[SFT-664]: Create Misc Integrations settings page

gustavs-gutmanis:
* adding "Other" integrations page
* creating `Google Tag Manager` integration
* creating `Javascript Test` integration
* creating `Honeypot` integration
* creating `Post Forwarding` integration
* removing any traces of previous honeypot logic
* removing previous GTM implementation


seandelaney (changes requested):
**Honeypot**

- [x] Honeypot is not saving if enabled is not selected. Fill in the fields and click save but the fields comes back as blank. If I fill in the fields, with enabled turned on and save, then saving works.
- [x] It would also be nice to default back to Honeypot tab again instead of the 1st tab.
- [x] Field handle should use handle middleware so we can save "Sean's Spam Prevention Field". This is not a valid input or body value and causes submission to be flagged as spam "Honeypot check failed"

**Javascript**

- [x] Does not save unless enabled is turned on.
- [x] Javascript Test field placeholder uses same value as Honeypot.
- [x] Input name field should handle middleware so we cannot save "what is this?" as a value.

image

`'input[name="what is this?"]'` is not valid HTML input and the POST payload includes "what is this?" and a blank value. 

- [x] Submission is also tagged as Spam with "Javascript Test failed".

**Google Tag Manager**

- [x] Event name field should prevent you from entering special characters like "sean's test event name". Is this also valid?

**GraphQL**

- [x] When I try to query a form that has no honeypot enabled, I get this error

image

seandelaney:
Within Settings > Other

The Other UI feels a bit disconnected from the other Integrations, that use the Craft UI / styles.

image

VS

image

kjmartens (approved):
Looks good

seandelaney:
@gustavs-gutmanis 

When I try to view Integrations tab, I now get this new error

image

seandelaney:
@gustavs-gutmanis @kjmartens 

The last issue I can find now is the honeypot fallback name showing `freeform_form_handle` instead of the value setup in settings. I will fix this in a separate PR so we can move this one forward.

seandelaney (approved):

PR #860, opened 2023-10-04
[SFT-677]: Implement Form Builder Defaults settings group

gustavs-gutmanis:
* adding default value CP view
* adding locked state to default values
* adding readers of default values to Properties
* adding locked state readers to Properties
* disabling locked items from being editable


kjmartens (changes requested):
Amazing!! 😍 

Noticing a few settings are not locking (being disabled/readonly):

- Form Type
- Formatting Template (disabled correctly in modal, NOT disabled in builder)
- Store Submitted Data for this Form (disabled correctly in modal, NOT disabled in builder)

There also needs to be a dropdown setting for Success Template (put it below _Success & Errors -> Return URL_):
`Success & Errors -> Success Template` (default not selected)

kjmartens (approved):

PR #861, opened 2023-10-05
[SFT-695]: 🔥 When saving an integration (non-singleton) in the Settings area it now triggers an `Undefined array key "class"` error

gustavs-gutmanis:
* fixing integrations


kjmartens (approved):
👌

Issue #783, opened 2023-08-23
Locale not working when using more than one Date & time field in the same form


pascalminator:
**Describe the bug or issue you're experiencing**

When I create multiple date & time fields in the same form, the first field is localised in french correctly, but when I create additional date & time fields they stay in english. I also tried to force the localisation by adding 'fr' in the config but it doesn't change anything.

**Steps to reproduce**

1. Use Craft localised in a different language than english (french for me)
2. Create a form and add multiple date & time fields
3. Preview your form in front-end and notice the first field being correctly localised and the additionnal ones incorectly displaying in english

**Expected behavior**

All date & time field should be in french

**Craft & Plugin Info (please complete the following information):**

- Craft Version: 4.4.17
- Freeform Version: 4.1.6
- Freeform Edition: Pro (trial because the website is not in prod yet)
- Fresh Install or Upgrade: Fresh install

kjmartens:
Sorry for the trouble @pascalminator.

I can duplicate this in my testing as well. I've noted it and we'll see if we can fix this shortly. 🙂

kjmartens:
Sorry for the long wait on this one @pascalminator. We were able to come up with a solution for this and it's now available in Freeform v4.1.11. 🙂

PR #875, opened 2023-10-11
[SFT-606] - Locale not working when using more than one Date & time field in the same form

seandelaney:
https://github.com/solspace/craft-freeform/issues/783

There was a bug with Flatpickr when calling the CSS/JS via a CDN and you had multiple date/time fields with the same locale on the same form.

The CDN version could not set the same locale multiple times, whether that was using `{ locale: 'fr' }` in the config or using `set('locale', 'fr')` option on the instance. Both approaches caused Flatpickr to throw an invalid locale error.

The bug does not exist when calling Flatpickr via Modules, using the latest version.


gustavs-gutmanis (commented):
Could you please check how much does the main JS artifact size increases when `flatpickr` is added as a dependency?

Previously we only loaded `flatpickr` when there are actual date fields in the form, but now it would be loaded always.

We had some clients complain that `flatpickr` is loaded even though they don't use it in their forms, so it would become a concern again. Is there any possibility at all to somehow fix the previous approach?

kjmartens:
Used https://github.com/solspace/craft-freeform/pull/897 instead

PR #878, opened 2023-10-12
[SFT-531]: Add Notices to top of Forms page

gustavs-gutmanis:
### Description
* adding notices to form dashboard
* adding error log counter in the dashboard

### Images
image


kjmartens (changes requested):
Gorgeous! Just needs a few tweaks:

1. The dismiss icon doesn't seem to do anything. Doesn't show any JS errors either.
2. The update notices seem to NOT use the `freeform_lock` and I could only get them to load when clearing CRAFT cache.
3. Can we force the order by type?
    - Critical
    - Warning
    - Info
    - New (which won't be used)
    - Error log at bottom (can stay the way you have it)

kjmartens (approved):

PR #882, opened 2023-10-12
[SFT-671] - Support ability to encrypt/decrypt submission data

seandelaney:
- Added new field property to encrypt field data.
- Added checks to show new encrypt field data property for Lite and Pro editions.
- Added new Encryption helper.
- Updated `Submission` element to encrypt submission fields on `beforeSave` if field can be encrypted and setting enabled.
- Updated `SubmissionQuery` to trigger decrypt submission fields on `afterPopulateElement` if field can be decrypted.
- Added new `EncryptionInterface` and `EncryptionTrait` to control which fields can be encrypted/decrypted.
- Fixed exporting submission data.
- Added new Input Field Type to better manage form controls


kjmartens (dismissed):
When I try viewing the submission data inside the CP, front end and "Updating Search Indexes" in Craft Queue, I get an error:
`base64_decode(): Argument #1 ($string) must be of type string, array given`

When attempting to submit a multipage form (not sure if it's the fact that it's multipage or something else), the form fails at the final submit and has this error:
```
Exception 'TypeError' with message 'openssl_encrypt(): Argument #1 ($data) must be of type string, array given'

in /YOINK/freeform5.c4/vendor/yiisoft/yii2/base/Security.php:218
```

On a single page form it appeared to work ok for submit.

packages/plugin/src/Bundles/Fields/Types/FieldTypesProvider.php
```diff
@@ -45,29 +44,6 @@ public function getSections(): array
...
-                            null,
-                            'Configuration',
-                            file_get_contents(__DIR__.'/../../../Fields/SectionIcons/gears.svg'),
-                            1
-                        ),
-                    ],
-                    $list,
-                );
-            }
-
```
gustavs-gutmanis: Why did you remove this?
There are times when no default section is preconfigured.

packages/plugin/src/Bundles/Submissions/CipherFieldsBundle.php
```diff
@@ -0,0 +1,91 @@
...
+            Submission::EVENT_ENCRYPT_FIELDS,
+            [$this, 'encryptFields']
+        );
+
+        Event::on(
+            Submission::class,
+            Submission::EVENT_DECRYPT_FIELDS,
+            [$this, 'decryptFields']
+        );
+    }
```
gustavs-gutmanis: With PHP8.0 you can use constructor arguments to define object properties. To avoid having to write boilerplate.

```suggestion
class CipherFieldsBundle extends FeatureBundle
{
    public function __construct(private CipherHelper $cipherHelper)
    {
        Event::on(
            Submission::class,
            Submission::EVENT_ENCRYPT_FIELDS,
            [$this, 'encryptFields']
        );

        Event::on(
            Submission::class,
            Submission::EVENT_DECRYPT_FIELDS,
            [$this, 'decryptFields']
        );
    }
```

packages/plugin/src/Bundles/Submissions/CipherFieldsBundle.php
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * @throws Exception
+     * @throws InvalidConfigException
+     * @throws \Exception
+     */
+    public function decryptFields(CipherEvent $event): void
+    {
+        if (!Freeform::getInstance()->isPro() || !\Craft::$app->getUser() || !PermissionHelper::checkPermission(Freeform::PERMISSION_SUBMISSIONS_ACCESS)) {
```
gustavs-gutmanis: We don't need to check for the user or submission access.
Also we should probably decrypt the values even for `Lite` editions.
Imagine a scenario where people roll back their edition only to find all of their fields encrypted and they're no longer able to read them.

packages/plugin/src/Elements/Db/SubmissionQuery.php
```diff
@@ -101,6 +103,16 @@ public function spamReason($value): self
...
 
+    public function afterPopulate(array $elements): array
+    {
+        foreach ($elements as $submission) {
+            $cipherEvent = new CipherEvent($submission);
+            Event::trigger(Submission::class, Submission::EVENT_DECRYPT_FIELDS, $cipherEvent);
+        }
+
+        return parent::afterPopulate($elements);
+    }
```
gustavs-gutmanis: There's an `SubmissionQuery::EVENT_AFTER_POPULATE_ELEMENT` and `SubmissionQuery::EVENT_AFTER_POPULATE_ELEMENTS` events. I would rather have our encryption bundle listen to those and perform the encrypt/decrypt there.

packages/plugin/src/Elements/Submission.php
```diff
@@ -135,6 +140,14 @@ public function __isset($name): bool
         return parent::__isset($name);
     }
 
+    public function beforeSave(bool $isNew): bool
+    {
+        $cipherEvent = new CipherEvent($this);
+        Event::trigger(self::class, self::EVENT_ENCRYPT_FIELDS, $cipherEvent);
+
+        return parent::beforeSave($isNew);
+    }
```
gustavs-gutmanis: Same here, there's a `Submission::EVENT_BEFORE_SAVE` event, which we can listen to, to perform encryption.

packages/plugin/src/Fields/AbstractField.php
```diff
@@ -92,7 +93,7 @@ abstract class AbstractField implements FieldInterface, IdentificatorInterface
         handle: 'attributes',
         label: 'Attributes',
         icon: __DIR__.'/SectionIcons/list.svg',
-        order: 999,
```
gustavs-gutmanis: `4` will definitely not work well with fields that have more than 4 sections.
It's `999` because it should always be at the very bottom.

packages/plugin/src/Fields/AbstractField.php
```diff
@@ -536,6 +550,16 @@ public function implements(string ...$interfaces): bool
         return false;
     }
 
+    public function isUseEncryption(): bool
```
gustavs-gutmanis: if we would use `protected bool $encrypted`, we could call the getter `isEncrypted()` instead. (Otherwise `PropertyAccessor` won't find this property.)

packages/plugin/src/Services/SubmissionsService.php
```diff
@@ -285,9 +285,13 @@ public function getAsArray(array $submissionIds): array
         $forms = Freeform::getInstance()->forms->getResolvedForms(['id' => $formIds]);
         foreach ($forms as $form) {
             $alias = 'fc'.$form->getId();
+            $fieldsArray = [];
+            foreach ($form->getLayout()->getFields() as $field) {
+                $fieldsArray[] = $field;
+            }
             $fields = array_map(
                 fn (FieldInterface $field) => $alias.'.[['.Submission::getFieldColumnName($field).']] as '.$field->getHandle(),
-                $form->getLayout()->getStorableFields()
```
gustavs-gutmanis, line 290: If you use `HTMLField` or any other field which doesn't store values in the DB, you will get a critical failure.

What's the idea behind this change?

packages/plugin/src/templates/_components/modals/export_csv.html
```diff
@@ -48,7 +48,7 @@ <h1>{{ 'Export data'|t('freeform') }}</h1>
             <div class="select">
                 <select class="select" name="form_id">
                     {% for form in forms %}
-                        <option value="{{ form.id }}"{{ (form.id == selectedFormId) ?? ' selected' }}>
+                        <option value="{{ form.id }}" {% if form.id == selectedFormId %}selected{% endif %}>
```
gustavs-gutmanis: Why this change?
I guess it should have been `{{ form.id == selectedFormId ? 'selected' }}`

packages/plugin/src/templates/_components/modals/export_csv.html
```diff
@@ -85,7 +85,7 @@ <h1>{{ 'Export data'|t('freeform') }}</h1>
                                    class="checkbox"
                                    name="export_fields[{{ form.id }}][{{ fieldId }}][checked]"
                                    value="1"
-                                   {{ isChecked ? 'checked' }}
+                                   {{ isChecked ?? 'checked' }}
```
gustavs-gutmanis: `??` makes `checked` output only when `isChecked` is `null`.
`?` is the first half of a terniary expression, which means it will output `checked` if `isChecked` evaluates to `true`.

It is a shorthand for
`{{ isChecked ? 'checked' : '' }}`
Twig lets us not write the `: ''` part, since it's useless here.

packages/plugin/src/controllers/export/QuickExportController.php
```diff
@@ -280,9 +281,25 @@ public function actionIndex()
...
+            foreach ($submission as &$field) {
+                if ($field) {
+                    $decryptedValue = \Craft::$app->getSecurity()->decryptByKey(base64_decode($field), $key);
+
+                    if ($decryptedValue) {
+                        $field = $decryptedValue;
+                    }
+                }
+            }
+        }
```
gustavs-gutmanis: Since we have the `Field` instance a couple of lines back, we can know if the value uses encryption or not, so we could avoid calling `decrypt` on all values.

It might be redundant, though, but this logic is used 3 times across the codebase and would be a prime candidate for being extracted somewhere, so as not to duplicate it.

packages/plugin/src/Fields/Interfaces/NoEncryptionInterface.php
```diff
@@ -0,0 +1,20 @@
...
+ * @copyright     Copyright (c) 2008-2022, Solspace, Inc.
+ *
+ * @see           https://docs.solspace.com/craft/freeform
+ *
+ * @license       https://docs.solspace.com/license-agreement
+ */
+
+namespace Solspace\Freeform\Fields\Interfaces;
+
+interface NoEncryptionInterface
```
gustavs-gutmanis: Lets make this into an `EncryptionInterface` and have it have just `isEncrypted()` method.

Then make an `EncryptionTrait` which holds the property and its definition as well as the getter.

Then we assign it to the fields which can be encrypted, like `Text`, `Options` etc. 
Remember - there are several base fields which are extended by lots of other fields, so we need to apply the interface and trait only to the base ones. (i.e. `Text` will cover a wast amount of other fields).

packages/plugin/src/Fields/AbstractField.php
```diff
@@ -127,6 +128,19 @@ abstract class AbstractField implements FieldInterface, IdentificatorInterface
     )]
     protected FieldAttributesCollection $attributes;
 
+    #[Section(
+        handle: 'advanced',
+        label: 'Advanced',
+        icon: __DIR__.'/SectionIcons/advanced.svg',
+        order: 5,
```
gustavs-gutmanis: Advanced sections should go below the `Attributes` section, so an order of `1000` here, I guess.

packages/plugin/src/Bundles/Fields/Types/FieldTypesProvider.php
gustavs-gutmanis: After reviewing the whole PR I see that this is now a manual entry everywhere instead of a generated default one.

Why this approach?
Before - we made sure there's always a general default section where every property that doesn't belong to a specific section falls in.
Now it's not only crazy amounts of work for you to manually type it everywhere, but it also introduces a possibility for human error - simply forgetting to enter a section.
If you don't enter one for a property, it will not show up anywhere in the builder anymore.

I'm guessing that you removed this because of the `advanced` section? If that's the case, we should think of a way to handle that case automatically, rather than with this manual approach.

packages/client/src/app/pages/forms/edit/builder/tabs/layout/property-editor/editors/fields/field-properties.tsx
```diff
@@ -75,6 +75,39 @@ export const FieldProperties: React.FC<{ uid: string }> = ({ uid }) => {
               property={property}
             />
           ))}
+          {handle === 'advanced' && (
```
gustavs-gutmanis: I'm thinking that this won't work if there's no `encrypt` property.
This has to be revised on my end.
For now leave it as it is here.

packages/client/src/app/components/form-controls/index.tsx
```diff
@@ -52,6 +53,10 @@ export const FormComponent: React.FC<Props> = ({
     return null;
   }
 
+  if (!config.editions.is(Edition.Pro) && flags?.includes(Edition.Pro)) {
+    return null;
+  }
```
gustavs-gutmanis: This should be covered already by the `PropertyProvider`.
It turns all `pro` fields into hidden ones if the edition isn't PRO.

packages/plugin/src/Bundles/Submissions/CipherFieldsBundle.php
```diff
@@ -0,0 +1,91 @@
...
+        if (!Freeform::getInstance()->isPro()) {
+            return;
+        }
+
+        $submission = $event->getSubmission();
+
+        $key = CipherHelper::getKey($submission->getForm());
+
+        foreach ($submission->getIterator() as $field) {
+            $value = $field->getValue();
```
gustavs-gutmanis: We should be checking for fields which have implemented the `EncryptionInterface` (which we have to make).

```php
$fields = $submission->getFieldCollection()->getList(EncryptionInterface::class);
```

And check if `$field->isEncrypted()` -> then do encryption

packages/plugin/src/Events/Submissions/CipherEvent.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Solspace\Freeform\Events\Submissions;
+
+use Solspace\Freeform\Elements\Submission;
+use Solspace\Freeform\Events\CancelableArrayableEvent;
+
+class CipherEvent extends CancelableArrayableEvent
```
gustavs-gutmanis: We can avoid this event it seems.

packages/plugin/src/Fields/AbstractField.php
```diff
@@ -127,6 +128,19 @@ abstract class AbstractField implements FieldInterface, IdentificatorInterface
...
+    )]
+    #[Flag(Flag::PRO)]
+    #[Input\Boolean(
+        label: 'Encrypt field data',
+        order: 7
+    )]
+    protected bool $encryption = false;
```
gustavs-gutmanis: We can crate an `EncryptionTrait` which has the definition of this property, but call it `$encrypted`, for better naming of the getter method. (helps with `PropertyAccessor`)

packages/plugin/src/Fields/AbstractField.php
```diff
@@ -536,6 +550,16 @@ public function implements(string ...$interfaces): bool
...
 
+    public function isUseEncryption(): bool
+    {
+        return $this->encryption;
+    }
+
+    public function canUseEncryption(): bool
+    {
+        return !$this instanceof NoEncryptionInterface;
+    }
```
gustavs-gutmanis: Lets go the inclusion route, rather than the exclusion one. Then we won't need this at all.

packages/plugin/src/Fields/BaseGeneratedOptionsField.php
```diff
@@ -15,6 +16,7 @@
  */
 abstract class BaseGeneratedOptionsField extends BaseOptionsField implements GeneratedOptionsInterface
 {
+    #[Section('configuration')]
```
gustavs-gutmanis: I have to say, this copy & paste approach here is a bad solution to a problem which doesn't exist.
Lets try and avoid such practices like the plague if at all possible.

packages/plugin/src/Fields/Implementations/Pro/GroupField.php
```diff
@@ -17,10 +19,14 @@
     typeShorthand: 'group',
     iconPath: __DIR__.'/../Icons/group.svg',
 )]
-class GroupField extends AbstractField implements NoStorageInterface, ExtraFieldInterface
+class GroupField extends AbstractField implements NoEncryptionInterface, NoStorageInterface, ExtraFieldInterface
```
gustavs-gutmanis: Lets add `EncryptionInterface` to all the fields which can be encrypted, rather than `NoEncryptionInterface` to the ones that can't.

There are some fields like Checkbox fields which values probably don't makes sense to encrypt, so we wouldn't be targeting all fields I guess.

packages/plugin/src/Fields/Implementations/Pro/GroupField.php
```diff
@@ -17,10 +19,14 @@
...
 {
     protected bool $required = false;
 
+    // Hides option in Pro edition
+    #[Flag('')]
+    protected bool $encryption = false;
```
gustavs-gutmanis: Can be removed

packages/plugin/src/controllers/export/ProfilesController.php
```diff
@@ -141,9 +159,23 @@ public function actionExport(): void
...
+            foreach ($submission as &$field) {
+                if ($field) {
+                    $decryptedValue = \Craft::$app->getSecurity()->decryptByKey(base64_decode($field), $key);
+
+                    if ($decryptedValue) {
+                        $field = $decryptedValue;
+                    }
+                }
+            }
+        }
```
gustavs-gutmanis: This logic tends to repeat in several places, it would be nice to extract it into a single point of entry, so that we duplicate code as minimally as possible.

gustavs-gutmanis (commented):
Adding some commentary.

packages/client/src/app/components/form-controls/index.tsx
seandelaney: I didn't know this.

packages/plugin/src/Services/SubmissionsService.php
seandelaney: ~`getStorableFields` doesn't exist anymore~

It does here. `$form->getLayout()->getFields()->getStorableFields()`

Updated

packages/plugin/src/templates/_components/modals/export_csv.html
seandelaney: I was getting a template error. Updated to `{{ form.id == selectedFormId ? 'selected' }}`

packages/client/src/app/components/form-controls/index.tsx
gustavs-gutmanis: That's cool, just mentioning it, since I know it's easy to miss this stuff.

seandelaney:
@kjmartens @gustavs-gutmanis 

I think we need to spend some time refactoring how exports work. Multiple export routes with multiple options.

seandelaney:
@kjmartens 

> When I try viewing the submission data inside the CP, front end and "Updating Search Indexes" in Craft Queue, I get an error: `base64_decode(): Argument #1 ($string) must be of type string, array given`
> 
> When attempting to submit a multipage form (not sure if it's the fact that it's multipage or something else), the form fails at the final submit and has this error:
> 
> ```
> Exception 'TypeError' with message 'openssl_encrypt(): Argument #1 ($data) must be of type string, array given'
> 
> in /YOINK/freeform5.c4/vendor/yiisoft/yii2/base/Security.php:218
> ```
> 
> On a single page form it appeared to work ok for submit.

I cannot replicate these issues in latest changes.  Can you please test again.

kjmartens (changes requested):
Array type fields seem to have issues with this feature. Affected fields are:

- Checkboxes
- Multiple Select
- File Upload
- File Upload Drag & Drop

1. When the form contains ANY fields that contain an array of options, when submitting the form (and that field has data), the form submit fails with an error:
`Exception 'TypeError' with message 'openssl_encrypt(): Argument #1 ($data) must be of type string, array given'  in /PATH/vendor/yiisoft/yii2/base/Security.php:218`
2. When the form contains ANY fields that contain an array of field data, it errors when loading the submission, even if it is NOT set to be encrypted.
```TypeError base64_decode(): Argument #1 ($string) must be of type string, array given```

**Review by kjmartens — DISMISSED** [review]

**Review by kjmartens — CHANGES_REQUESTED** [review]
Much closer to working correctly now. However... 😅 

- Exporting via Quick Export and Export Profiles does not correctly decrypt the array fields (Checkboxes, Multiple Select, File Upload, File Upload Drag & Drop.

**Review by kjmartens — APPROVED** [review]

**Comment by kjmartens** [comment]
Seems good now. We should just wait on @gustavs-gutmanis's review when he gets back. 🙂

**Review by kjmartens — DISMISSED** [review]
A couple more issues I found with editions:

1. Encryption is only supposed to be available for Pro, but appears on Lite as well (correctly hidden on Express).
2. Decryption should happen regardless of edition. For example, if I was using Pro and encrypted some data and then later on downgraded to Lite (or Express) edition, I should still be able to access/view the decrypted submission data. Currently, it shows old data as encrypted still.

**Review by kjmartens — CHANGES_REQUESTED** [review]
Latest Changes:

- Encryption still shows for `Lite` when it should only appear for Pro (correctly hidden on Express).

**Review by kjmartens — APPROVED** [review]

File: packages/plugin/src/Fields/Implementations/Pro/PasswordField.php [hunk 24]
```diff
@@ -26,6 +27,9 @@
 )]
 class PasswordField extends TextField implements NoStorageInterface, ExtraFieldInterface, RememberPostedValueInterface
 {
+    #[Input\Hidden]
```
gustavs-gutmanis, line 30: Does this need `#[Input\Hidden]`?
Does react need to know about its hidden state somehow?

If not, I think we can remove the attribute.

packages/plugin/src/Library/Helpers/EncryptionHelper.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * @throws Exception
+     * @throws InvalidConfigException
+     */
+    public static function encrypt(string $key, mixed $value): mixed
+    {
+        if (\is_array($value)) {
+            foreach ($value as &$element) {
```
gustavs-gutmanis: Why are we iterating over references here?
We don't seem to be changing the original at all.

[Here's an explanation on why you should call `unset()`](https://www.php.net/manual/en/control-structures.foreach.php) on the value after a loop that uses a reference value.

gustavs-gutmanis: Also - maybe it makes sense to `json_encode` the values before encrypting and `json_decode` them on decrypt?

packages/plugin/src/Library/Helpers/EncryptionHelper.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * @throws Exception
+     * @throws InvalidConfigException
+     */
+    public static function decryptExportData(string $key, array $data): array
+    {
+        foreach ($data as &$row) {
+            foreach ($row as &$value) {
```
gustavs-gutmanis: I would advise against using reference values when looping, it's easy to miss that breaking thing I linked to you previously.

packages/plugin/src/Library/Helpers/EncryptionHelper.php
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        return $data;
+    }
+
+    /**
+     * @throws InvalidConfigException
+     * @throws Exception
+     */
+    public static function encryptByKey(string $key, mixed $value): string
```
gustavs-gutmanis: I think we can safely do `string $value` in the arguments, since it must always be a string anyway.

Or better yet, keep it as `mixed` and do `json_encode` here for the value, and adjust the decrypt accordingly.

packages/plugin/src/controllers/SubmissionsController.php
```diff
@@ -78,25 +79,32 @@ public function actionExport()
         $submissionIds = \Craft::$app->request->post('submissionIds');
         $submissionIds = explode(',', $submissionIds);
 
-        $submissions = $this->getSubmissionsService()->getAsArray($submissionIds);
+        $form = null;
```
gustavs-gutmanis: It seems like this variable is redundant, because it will always be defined in line 93.

packages/plugin/src/controllers/export/ProfilesController.php
```diff
@@ -123,6 +129,14 @@ public function actionDelete(): Response
...
     }
 
+    /**
+     * @throws InvalidConfigException
+     * @throws FreeformException
+     * @throws HttpException
+     * @throws ForbiddenHttpException
+     * @throws BadRequestHttpException
+     * @throws Exception
+     */
```
gustavs-gutmanis: Keeping these PHPDoc tags correct is an impossible task, because each time something in the chain changes, they become obsolete or outdated, so there's no real benefit of writing them. I would just disable the specific inspection in your JetBrains IDE.

gustavs-gutmanis (commented):
Looking real nice now,

Just a couple of more questions before the approval.

packages/plugin/src/Fields/Implementations/Pro/PasswordField.php
seandelaney: @gustavs-gutmanis This is hiding the Encrypt field data label/checkbox. Because Password extends Text, Encrypt field data was still showing as 1) its within Pro edition and 2) inheritances EncryptionInterface from Text (which cannot be removed).

gustavs-gutmanis: Yeah I thought so, but if you override the variable definition, woudn't the attribute be unset as well?
If not, than this is the way to go, for sure, but I was just wondering.

packages/plugin/src/Library/Helpers/EncryptionHelper.php
seandelaney: We need to loop over certain field values as the field types such as table are saving values as an array in the DB itself. We cannot json_encode/json_decode the whole value for these multiple value fields as then we'd need to refactor how these field values are saving/used, with a ripple effect across the whole app.

packages/plugin/src/Library/Helpers/EncryptionHelper.php
seandelaney: Its to allow field types like table, checkboxes etc that save values as arrays to work.

seandelaney: It cant be a string as then tables etc wont work.

packages/plugin/src/Library/Helpers/EncryptionHelper.php
gustavs-gutmanis: The DB and checkbox fields store their array values as a JSON string in the DB, it is converted to and from arrays when loading the data via records, etc.

And the encrypted value is a string anyway, so maybe it is possible?
If a value is encrypted, it's 100% an encoded JSON. So even a single string value would be JSON encoded as `"somevalue"` when decrypted.

packages/plugin/src/Library/Helpers/EncryptionHelper.php
seandelaney: Would you create a new array instead ?

gustavs-gutmanis: Good question.
I guess this would conserve more memory on larger datasets.
Lets keep it like this.

A quick side question - is each value encrypted and decrypted separately in an array dataset?

packages/plugin/src/Fields/Implementations/Pro/PasswordField.php
seandelaney: TBH, I'm not 100% happy with this as a solution. Setting encrypted to false alone doesn't hide it. Unless you know why, a dev would be very confused. What do you think about making Password not extend Text?

gustavs-gutmanis:
I guess it's better to have this attribute than to get rid of the inheritance.

--
Gustavs Gutmanis

gustavs-gutmanis (approved):
Looks great, thanks!

PR #883, opened 2023-10-14
[SFT-641]: Implement Express edition

gustavs-gutmanis:
* Limit 1 form
* Limit 20 fields
* No elements/predefined feeders
* No Conditional email notifications
* No User Select email notifications
* No Set a Close Date per Form
* No multipage forms
* No javascript test
* No element field types
* No GraphQL support


kjmartens (changes requested):
My bad, I forgot a couple of items to be excluded in Express edition:

- Email Alerts
- Digest Email Notifications
Screenshot 2023-10-14 at 5 29 49 PM

- POST Forwarding (this was actually an issue where the Lite edition is also including it... it's a PRO only feature so it needs to be fixed for both Express and Lite).
- In this branch only, the form builder settings tab is all squished... not sure what happened.
Screenshot 2023-10-14 at 5 41 53 PM

kjmartens (approved):

PR #885, opened 2023-10-17
[SFT-718]: When using Digest feature, clearing the Craft cache will fail on an error...

gustavs-gutmanis:
### Description
* fixing digest generating incorrect `RecipientsCollection`
* removing unused Command `Notifications`
* removing unused `WeeklyDigestJob`


kjmartens (approved):
Fixed now

PR #886, opened 2023-10-17
[SFT-716]: Let's remove tool-tips from the builder settings (for description)

gustavs-gutmanis:
### Description
* removing tooltips from property editors


kjmartens (approved):
Approving this, but noting another issue that popped up...

Issue #888, opened 2023-10-18
GraphQL submission mutation has empty fields


paulnzime:
**Describe the bug or issue you're experiencing**

Unable to submit mutation via GraphQL. It submits and creates the record in the CMS but the fields are blank.

**Steps to reproduce**

1. Enable the public schema and tick the boxes for submission to the form
2. Submit GraphQL mutation using Postman or similar, not the Craft GraphQL Explorer.
3. Form submission is a success but check in the CMS and the fields won't be saved

**Craft & Plugin Info (please complete the following information):**

- Craft Version: 4.5.6.1
- Freeform Version: 4.1.10
- Freeform Edition: Lite
- Fresh Install or Upgrade: Fresh

**Additional context**

Inspecting the code I have found this if statement in `packages/plugin/src/Bundles/Form/Context/Request/GraphQLContext.php`:

`if ($request->getHeaders()->get('x-craft-gql-schema') || $request->getHeaders()->get('X-Craft-Solspace-Freeform-Mode')) {`

This wraps around the part of the code that seems to turn the request data into the form submission.
In Postman if I now set a header value matching one of these, the submission works as expected.

Why is this a requirement? I would not expect to have to declare a header on the GraphQL mutation.

kjmartens:
Sorry for trouble you're experiencing with this @paulnzime. We reviewed this and were able to remove the requirement for this in Freeform v4.1.11. 🙂 

Can you try it out and see if that resolves the issue for you? 🙂

PR #894, opened 2023-10-20
[SFT-725] - Removed Custom Header requirement for GraphQL submission …

seandelaney:
Headless approaches using React/Next/Vue no longer have to pass a specific Freeform header in GraphQL requests for submission mutations to work.

Resolves issue https://github.com/solspace/craft-freeform/issues/888


gustavs-gutmanis (approved):
👍🏻

PR #896, opened 2023-10-24
[SFT-606]: Locale not working when using more than one Date & time field in the same form

gustavs-gutmanis:
### Description
* fixing datepicker locale assignment on consecutive datepicker elements


kjmartens (approved):
Tested and works

PR #903, opened 2023-11-01
[SFT-720] - Fixed if you remove a field from a form, it does not automatically remove it for any Export Profiles for that form, causing an error

seandelaney:


kjmartens (approved):
Seems to resolve the issue

gustavs-gutmanis (approved):
Nice, thanks! 👍🏻

PR #908, opened 2023-11-07
[SFT-738] - Updated GraphQL mutations support and fixed a bug with ActiveCampaign integration

seandelaney:


packages/plugin/src/Fields/AbstractField.php
```diff
@@ -433,6 +433,11 @@ public function getIdAttribute(): string
         return $this->parameters->fieldIdPrefix.$attribute;
     }
 
+    public function getContentGqlHandle(): ?string
+    {
+        return $this->getHandle();
+    }
+
```
gustavs-gutmanis, line 440: If this is just an alias for `->getHandle()` do we really need to rewrite so many places?

seandelaney: Probably not right now. Being proactive. I copied this pattern over from v4 fixes where some fields needed to use GQL specific handles and thought it was a good idea to keep same pattern.

gustavs-gutmanis (approved):

PR #914, opened 2023-11-09
[SFT-746]: Success and error messages are not escaped for use in HTML attributes

gustavs-gutmanis:
* pulling out `StringHelper` from `Commons` as a built-in library in FF
* fixing attributes not being escaped properly when compiled


kjmartens (approved):
Excellent 👌

PR #916, opened 2023-11-09
[SFT-745]: The `freeform_lock` isn't updating on a customer's site, causing other issues

gustavs-gutmanis:
### Description

* moving away from `Element::delete()` to a direct DB SQL DELETE expression that removes elements by ID's.


kjmartens (approved):
Seems to work for me

PR #921, opened 2023-11-22
[SFT-722] - Added fallback value when field label is blank

seandelaney:
When field label is set to blank, we attempt to use field label from the blueprint / freeform_fields table.
If no field is found, we fallback to using the field handle.


kjmartens (changes requested):
Looks really good! However, one change must be made (for legacy)... if someone has an empty label for a field, it should **NOT** provide a fallback label for their front end form. There are definitely some customers currently relying on the old behavior (for front end).

kjmartens (approved):
Looks good! Thank you

gustavs-gutmanis (approved):
Looks great!

PR #928, opened 2023-11-30
[SFT-721]: Stripe Integration

gustavs-gutmanis:
### Description
* adding stripe integration as a feature bundle
* adding quickSave to form js plugin
* adding events to allow customizing submission fields and values


kjmartens (approved):
Issues noted as tasks in Jira

PR #931, opened 2023-12-04
[SFT-760]: When using Dynamic Amount for Stripe AND the beginning value is empty (or 0), it breaks the Stripe element

gustavs-gutmanis:
### Description
* fixing stripe script crashing on empty amount field values


kjmartens (approved):
Looks good

PR #939, opened 2023-12-13
fix(SFT-700): revised diagnostics page

thejahid:
Freeform includes a self-diagnostics page that may help customers track down potential issues with their setup and allow them to easily share diagnostic results with the software Support team to assist them faster.


packages/plugin/src/Library/DataObjects/Summary/Statistics/Spam.php
```diff
@@ -39,4 +39,10 @@ class Spam
 
     /** @var string */
     public $captchaType = '';
+
+    /** @var bool */
+    public $bypassSpamCheckOnLoggedInUsers = false;
```
gustavs-gutmanis: If you notice any places where the property type isn't defined - it's very welcome to specify them.

```suggestion
    public bool $bypassSpamCheckOnLoggedInUsers = false;
```

packages/plugin/src/Services/DiagnosticsService.php
```diff
@@ -170,306 +237,273 @@ public function getFreeformStats()
...
+                'Form types: <b>{{ value|length }}</b>',
+                \count($formTypes)
+            );
+        }
+
+        return $diagnosticItems;
     }
 
-    public function getFreeformChecks()
+    public function getFreeformIntegrations()
```
gustavs-gutmanis: Lets strive to always define return types of functions.

```suggestion
    /**
     * @return DiagnosticItem[]
     */
    public function getFreeformIntegrations(): array
```

In this case, we would also need a @DocType since the `array` return type doesn't encompass the type of elements inside the array. Otherwise the @DocType is unnecessary.

packages/plugin/src/Services/DiagnosticsService.php
```diff
@@ -170,306 +237,273 @@ public function getFreeformStats()
...
-                            fn ($value) => $value,
-                            '',
-                            'It is beneficial for most sites to have this enabled to catch false positives and/or look for patterns with spam.'
-                        ),
-                    ],
-                    fn ($value) => $value ? DiagnosticItem::COLOR_PASS : DiagnosticItem::COLOR_BASE
-                ),
+        foreach ($integrations as $integration) {
+            $name = $integration['name'];
+            $version = isset($integration['version']) && '' !== $integration['version'] ? "({$integration['version']})" : '';
```
gustavs-gutmanis: Rule of thumb here would be to always define a variable when having to lookup and use an array key more than once.

But in this case, you can just define the variable once, and then modify itself if the value is present;

Also, there's no need to do a `'' !==` check, since it will either be empty or it won't.

```suggestion
            $version = $integration['version'];
            $version = $version ? "({$version})" : ''
```

packages/plugin/src/Services/DiagnosticsService.php
```diff
@@ -170,306 +237,273 @@ public function getFreeformStats()
...
+            $count = $integration['count'];
+
+            $plural = 1 != $count ? 's' : '';
```
gustavs-gutmanis: I'm not 100% sure, we can wait for @kjmartens input, but I think there's no benefit for us to use plurals at all here, because translating it into different languages might not work with plurals, since some languages have a vastly different approach to plurals.

packages/plugin/src/Services/DiagnosticsService.php
```diff
@@ -572,4 +598,32 @@ private function getSummary(): InstallSummary
...
+
+    private function getIntegrationCount(): array
+    {
+        $integrations = (new Query())
+            ->select(['forms_integrations.id', 'forms_integrations.integrationId', 'forms_integrations.formId', 'integrations.class'])
+            ->from(FormIntegrationRecord::TABLE.' forms_integrations')
+            ->innerJoin(IntegrationRecord::TABLE.' integrations', 'integrations.id = forms_integrations.integrationId')
+            ->where(['forms_integrations.enabled' => true])
+            ->all()
+        ;
```
gustavs-gutmanis: Personally I think having full table name aliases make the queries less readable, rather than more readable.

So aliasing tables with initials I think makes more sense.

Instead of `forms_integrations.id` using `fi.id`

packages/plugin/src/Services/DiagnosticsService.php
```diff
@@ -572,4 +598,32 @@ private function getSummary(): InstallSummary
...
+        $integrationsByForm = [];
+
+        foreach ($integrations as $integration) {
+            $id = $integration['integrationId'];
+            if (!isset($integrationsByForm[$id])) {
+                $integrationsByForm[$id] = [
+                    'name' => $this->integrationTypeProvider->getTypeDefinition($integration['class'])->name,
+                    'version' => $this->integrationTypeProvider->getTypeDefinition($integration['class'])->version,
+                    'count' => 0,
+                ];
```
gustavs-gutmanis: Since we're using that huge path of a parameter more than once, it makes for cleaner code to extract it into a variable.

```suggestion
                $type = $this->integrationTypeProvider->getTypeDefinition($integration['class']);
                $integrationsByForm[$id] = [
                    'name' => $type->name,
                    'version' => $type->version,
                    'count' => 0,
                ];
```

And then, to prevent the `version` from being `null`, we can default it to an empty string.


```
$type = $this->integrationTypeProvider->getTypeDefinition($integration['class']);
$integrationsByForm[$id] = [
    'name' => $type->name,
    'version' => $type->version ?? '',
    'count' => 0,
];
```

packages/plugin/src/Services/Form/FieldsService.php
```diff
@@ -136,6 +138,21 @@ public function getAllFieldCount(): int
         return FormFieldRecord::find()->count();
     }
 
+    public function getFavoriteFieldCount(): int
+    {
+        return FavoriteFieldRecord::find()->count();
+    }
```
gustavs-gutmanis, line 143: This makes sense here 👍🏻

packages/plugin/src/Services/Form/FieldsService.php
```diff
@@ -136,6 +138,21 @@ public function getAllFieldCount(): int
...
+
+    public function getFormsWithStripeFieldCount(): int
+    {
+        return FormFieldRecord::find()
+            ->select('formId')
+            ->distinct()
+            ->where(['type' => StripeField::class])
+            ->count()
+        ;
+    }
```
gustavs-gutmanis: This doesn't make sense here.

First of all, this is very Stripe centric, meaning that only the Stripe  integration should be handling this logic, since Freeform should know nothing about this otherwise.

But for posterity - since it will only be used in the diagnostics page, we can have this logic inside the Diagnostics Service.

packages/plugin/src/Services/FormsService.php
```diff
@@ -124,6 +125,16 @@ public function getAllFormIds(): array
...
 
+    public function getAllRegularFormIds(): array
+    {
+        return $this
+            ->getFormQuery()
+            ->select('id')
+            ->where(['type' => Regular::class])
+            ->column()
+        ;
+    }
```
gustavs-gutmanis: I would modify this to broaden the use case.
Expose the method for fetching form ID's but let the dev specify a `$type` if they want, which will then modify the query to filter out a specific type.

Then in the place where you use this, pass in the `Regular::class` as the function argument.

```suggestion
    public function getAllFormIds(?string $type = null): array
    {
        $query = $this>getFormQuery()>select('id');
        if (null !== $type) {
            $query->where(['type' => $type]);
        }
        
        return $query->column();
    }
```

packages/plugin/src/templates/settings/_diagnostics.twig
gustavs-gutmanis, line 1: I have made several changes to the template to make it more readable and have less conditionals in it.

Templates should try to have as little logic inside them as possible, so I moved the checks out into the `DiagnosticItem` object instead, to make rendering templates easier and cleaner. Please see the changes I've made.

gustavs-gutmanis (changes requested):
Great job.
Please see my comments, there's a couple of things that I fixed, and there's a couple that I left for you to fix.

packages/plugin/src/templates/settings/_diagnostics.twig
thejahid: There is a {{ key }} that prints the array key, such as 1, 2, 3. Do we need to retain it in the HTML and apply something like 'sr-only' as in Tailwind?

gustavs-gutmanis: I don't think we need it. I see you already removed it, so that's good.

kjmartens (changes requested):
Just noticing a couple minor issues:

1. When both conditions apply for the Memory Limit, just the warning one should show, not both. So I guess the blue suggestion conditional might work better if it only shows when the memory limit is between 256M-511M (rather than below 512M).
Screenshot 2023-12-14 at 8 05 01 AM

2. The version numbers/type for reCAPTCHA and hCaptcha need to be handled a bit differently. They are single integrations where the user can select a version within it. You'll need to sniff out which version they chose for it, e.g. `reCAPTCHA (v2 Invisible)`
Screenshot 2023-12-14 at 8 06 10 AM

packages/plugin/src/Services/DiagnosticsService.php
kjmartens: @gustavs-gutmanis true, but I am also a sucker for good pluralization 😅

gustavs-gutmanis:
@kjmartens why wouldn't we show the suggestion along with the error?
They might opt to increase it to 512M rather than just 256M

packages/plugin/src/Services/DiagnosticsService.php
```diff
@@ -117,71 +117,48 @@ function ($value) {
...
-                                    $multiplier = 1024 ** 3;
-
-                                    break;
-                            }
+                            $multiplier = match ($measurement) {
+                                'k' => 1024,
+                                'm' => 1024 ** 2,
+                                'g' => 1024 ** 3,
+                                default => 1,
+                            };
```
gustavs-gutmanis, line 132: Gorgeous! 👌🏻

gustavs-gutmanis (approved):

kjmartens:
> @kjmartens why wouldn't we show the suggestion along with the error? They might opt to increase it to 512M rather than just 256M

@gustavs-gutmanis True... perhaps I could just change the wording for the warning, because otherwise it looks a little clumsy with both notices showing. 🙂

kjmartens (approved):

PR #942, opened 2023-12-19
[SFT-604] Add Created and Updated users

aaronlinley:


packages/client/src/app/pages/forms/edit/builder/tabs/form-settings/settings.sidebar.tsx
```diff
@@ -26,6 +29,11 @@ export const SettingsSidebar: React.FC = () => {
     return null;
   }
 
+  const { data: ownership } = useQueryFormOwnership(Number(formId || 0));
+  if (!ownership) {
+    return null;
+  }
```
gustavs-gutmanis: Seems like none of the settings will be rendered until this is fetched.
I think it's better to show the settings with the loader and display this after it's loaded.

Also this would be a good candidate for its own component which handles fetching and rendering inside itself.

packages/client/src/app/pages/forms/edit/builder/tabs/form-settings/settings.sidebar.tsx
```diff
@@ -70,6 +78,15 @@ export const SettingsSidebar: React.FC = () => {
           ))
         )}
       </SectionWrapper>
+      <SidebarSeperator />
+        <SectionWrapper>
+          <SidebarMeta>
+            Created by <SidebarMetaUserLink href="#">Admin</SidebarMetaUserLink> at:<br/> { ownership.created.datetime }
```
gustavs-gutmanis: We will need to translate these values.
There's a `translate()` function used throughout the project, and it allows for placeholder replacement.

packages/client/src/app/pages/forms/edit/builder/tabs/form-settings/settings.sidebar.tsx
```diff
@@ -3,7 +3,7 @@ import { useSelector } from 'react-redux';
 import { useNavigate, useParams } from 'react-router-dom';
 import { Sidebar } from '@components/layout/sidebar/sidebar';
 import { formSelectors } from '@editor/store/slices/form/form.selectors';
-import { useQueryFormSettings } from '@ff-client/queries/forms';
+import {useQueryFormOwnership, useQueryFormSettings} from '@ff-client/queries/forms';
```
gustavs-gutmanis: Be sure to either set your IDE to run lint or prettier automatically, or run the `npm run lint:fix` command. It should automatically format all your changes based on the eslint rules.

packages/client/src/types/forms.ts
```diff
@@ -36,3 +36,14 @@ export type FormSettingNamespace = {
   sections: Section[];
   properties: Property[];
 };
+
+export type FormOwnership = {
+  created: {
+    datetime: string;
+    user: {};
```
gustavs-gutmanis: I'm guessing `user` is going to be an ID and a name, so an object of 
```ts
{
  id: number;
  name: string;
}
```

Also both `created` and `updated` are of identical types, so we could define a single type before these, without exporting it, and have these two reuse it.

packages/plugin/src/controllers/api/FormsController.php
```diff
@@ -75,6 +75,37 @@ public function actionSort(): Response
         return $this->response;
     }
 
+    public function actionOwnership(int $id): Response
```
gustavs-gutmanis: When you open up the form builder for a specific form, it loads a serialized `Form` object from `admin/freeform/api/forms/[id]`.
Since you are adding two getter methods for `createdBy` and `updatedBy`, they should also be included in the serialization.

Please check the response from the endpoint, it might be that you already have this data fetched along with the form by default.

This is quite a small dataset which would benefit from being loaded with the rest of form settings, since each request prolongs the time in which the builder loads.

packages/plugin/src/migrations/m231219_105754_AddUsersToFormsTable.php
```diff
@@ -0,0 +1,69 @@
...
+class m231219_105754_AddUsersToFormsTable extends Migration
+{
+    /**
+     * @inheritdoc
+     */
+    public function safeUp(): bool
+    {
+        $this->addColumn(
+            '{{%freeform_forms}}',
+            'createdId',
```
gustavs-gutmanis: `createdId` does not really portray what the ID stands for, could you please name it as `createdByUserId` and `updatedByUserId` instead, please?

Also, please update the references to these throughout the code as well.

packages/plugin/src/migrations/m231219_105754_AddUsersToFormsTable.php
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Solspace\Freeform\migrations;
+
+use Craft;
+use craft\db\Migration;
+
+/**
+ * m231219_105754_AddUsersToFormsTable migration.
+ */
+class m231219_105754_AddUsersToFormsTable extends Migration
```
gustavs-gutmanis, line 11: Because of how Craft CMS is built, it's not enough to make a migration, you also need to clone this very same thing inside the `Install.php` migration as well.

I know... it's not optimal, but it is what it is.

packages/client/src/app/pages/forms/edit/builder/tabs/form-settings/settings.ownership.tsx
```diff
@@ -0,0 +1,51 @@
...
+import {useSelector} from "react-redux";
+
+export const SettingsOwnership: React.FC = ({ }) => {
+	const { ownership } = useSelector(formSelectors.current);
+
+	if (!ownership) {
+		return null;
+	}
+
+	console.log(ownership);
```
kjmartens: Just noting that the console log should be removed.

kjmartens (changes requested):
Looks really good. I tested upgrade as well as fresh install. Behavior seems correct. One improvement, however, would be to remove the link to other users if they don't have the `Edit → Edit users` Craft permission. Currently it just links to `#` which isn't the greatest. 🙂

packages/plugin/src/Bundles/Transformers/Builder/Form/FormTransformer.php
```diff
@@ -91,4 +93,45 @@ private function decorateWithSubmissionStatistics(array $forms): array
 
         return $forms;
     }
+
+    private function getOwnership(Form $form): array
+    {
+        $createdByUserUrl = "#";
+        $updatedByUserUrl = "#";
```
gustavs-gutmanis: These need to be null if no users present

packages/plugin/src/Bundles/Transformers/Builder/Form/FormTransformer.php
```diff
@@ -91,4 +93,45 @@ private function decorateWithSubmissionStatistics(array $forms): array
...
+        $currentUser = \Craft::$app->getUser()->getIdentity();
+        $ownership = [
+            'created' => [
+                'datetime' => $form->getDateCreated()->format('n/j/Y, g:i A')
```
gustavs-gutmanis: I would suggest sending a unix timestamp and using `date-fns` in react to format the date using local time formatting options instead.

gustavs-gutmanis (changes requested):
Stellar!
Just two more small updates and we're all set here.

aaronlinley:
@gustavs-gutmanis @kjmartens All PR feedback has now been addressed

kjmartens (approved):
Looks good to me 👌

kjmartens (changes requested):
One issue I noticed is that the time is not correctly localized to me. I am `GTM-06` and it's currently `12:00 PM` here, but the time saved shows `01/05/2024, 6:00 PM`. When I change my timezone to another one, it changes with it, but they are wrong, too. Is it possible the timezone offset is "backwards" or something?

Also, should the date and time formatting not respect the Craft localization? For example, if I switch my language to `Russian`, it shows up the same (with `am`/`pm` etc), but in Craft Entries it switches to `05.01.2024, 13:09`. It isn't a massive deal, but my assumption was that with the recent changes you made, this was the intention? 🙂

aaronlinley:
Yeah, I can see some issues with the localization. It's not quite working the way I thought it would when using the `date-fns` formatting that @gustavs-gutmanis mentioned. I need to look into it a little further

kjmartens:
> Yeah, I can see some issues with the localization. It's not quite working the way I thought it would when using the `date-fns` formatting that @gustavs-gutmanis mentioned. I need to look into it a little further

@gustavs-gutmanis can you meet with @aaronlinley to let him know which adjustments are necessary to resolve this? 🙂

gustavs-gutmanis:
I can see `date-fns` doesn't deal with timezones. If you want timezones, we must install the `date-fns-tz` package, but even that won't solve this, because it doesn't infer the timezone from the browser, like Date does.

You could try using the [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) to achieve this.

Or as an alternative - use Craft's date localization methods to send a parsed date string from the server already.

aaronlinley:
@kjmartens @gustavs-gutmanis I've updated the datetime handling to be more reliable now. It sends a ISO string from the server to the frontend, and uses `date-fns-tz` to format it in the current user's timezone. This displays the date correctly in the current timezone, and updates the value when the timezone is changed via Craft settings (e.g. GMT+1 displays the time as 1 hour earlier than GMT).

kjmartens (changes requested):
Hmm, I am still seeing no difference.

- Times are still +6hrs ahead for me (e.g. should say 11:00am for me, but says 5:00pm). My timezone is GMT-6.
- When I switch to a European locale for my user, dates/times are not formatted any differently (e.g. uses 12hr clock instead of 24hr clock).
- I am getting a new issue... about 30% of the time, a manual page refresh in the builder doesn't show the builder and console logs the following errors:

```
vendor.js?v=1704819269:2 TypeError: Cannot read properties of undefined (reading 'created')
    at t.SettingsOwnership (client.js?v=1704819269:1:259266)

vendor.js?v=1704819269:2 Uncaught TypeError: Cannot read properties of undefined (reading 'created')
    at t.SettingsOwnership (client.js?v=1704819269:1:259266)
```

The part of the JS file it refers to is:
`c.utcToZonedTime)(new Date(e.created.datetime), t)`

gustavs-gutmanis (approved):
Looking good! 👍🏻

kjmartens (approved):
Wonderful

PR #946, opened 2024-01-01
[SFT-783]: Field Group Manager 

kjmartens:


packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/modal/modal.styles.ts
```diff
@@ -0,0 +1,178 @@
...
+    content: ${({ $empty }) => `"${$empty}"`};
+    display: block;
+  }
+`;
+
+FieldTypes.defaultProps = {
+  $empty: 'Drag and drop any field here',
+};
+
+export const UHFieldWrapper = styled.div`
```
gustavs-gutmanis, line 116: What does `UH` stand for?

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/modal/model.operations.ts
```diff
@@ -0,0 +1,170 @@
...
+};
+
+const mapGroupIdsToGrouped = (
+  groupIds: string[],
+  initialState: Group,
+  fieldListRefs: React.MutableRefObject<FieldListRefs>
+): GroupData[] => {
+  return groupIds.map((id) => {
+    const group = initialState.groups?.grouped.find((item) => item.uid === id);
+    const foundGroupEl = fieldListRefs.current[id];
```
gustavs-gutmanis: We should avoid touching actual HTML Elements and their attributes.
If it makes it easier, we can just fetch the field type classes directly within the request for groups.

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/modal/modal.tsx
```diff
@@ -0,0 +1,236 @@
...
+  const [loaded, setLoaded] = useState(false);
+
+  const fieldListRefs = useRef<FieldListRefs>({});
+  const {
+    addGroup,
+    removeGroup,
+    updateGroupInfo,
+    addItemToUnassigned,
+    syncFormRefs,
+  } = useGroupOperations(state, setState, fieldListRefs);
```
gustavs-gutmanis: The functions inside there should not know about setState, and be just pure functions that receive the current state, the thing that has to be changed, and output a new state, and that should then be persisted via setState here. 

As I mentioned elsewhere - I think my idea about using array indexes for the grouped/ungrouped response was a bad idea, and it seems like it overly complicated things for you here. Lets refactor that part and make it easier for you to deal with this data

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/modal/modal.tsx
```diff
@@ -0,0 +1,236 @@
...
+        ...groups,
+        unassigned: unassignedTypes,
+      },
+    });
+
+    setLoaded(true);
+  }, [data, loaded]);
+
+  useEffect(() => {
+    initializeSortable(fieldListRefs, state);
```
gustavs-gutmanis: This will re-initialize sortable each time anything changes, is  that expected?
I noticed that the sortable behaviour is pretty laggy and slow, this might be the cause.

Does the state changes while you're dragging it? That would explain the issue.

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/modal/model.list-item.tsx
```diff
@@ -0,0 +1,42 @@
...
+  addItemToUnassigned,
+}) => {
+  const fieldType = useFieldType(typeClass);
+  const { name, icon } = fieldType;
+
+  const fieldItemRef = useRef<HTMLDivElement>(null);
+  const hovering = useHover(fieldItemRef);
+
+  const removeFieldItem = (): void => {
+    fieldItemRef.current?.parentNode?.removeChild(fieldItemRef.current);
```
gustavs-gutmanis: We should definitely never touch DOM, this can result in unforeseen issues.
This should be entirely handled by sortable un state updating.

Rendering items where they are supposed to be rendered based on state.

gustavs-gutmanis (changes requested):
Great work! 

We need to change our strategy a bit, otherwise it's making code too complex, please read my comments.

kjmartens:
Looks good to me, @thejahid 👍

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/base-fields.tsx
```diff
@@ -1,37 +1,89 @@
...
+            </>
+          )}
+        </GroupWrapper>
       ))}
-    </FieldGroup>
+
+      {groupsData?.types && (
+        <List>
+          {groupsData?.types.map((type) => {
+            const fieldType = fieldTypesData.find(
```
gustavs-gutmanis: There's hook for this, which would make your code cleaner
`useFieldTypeSearch` in `@ff-client/queries/field-types`.

```ts
const findType = useFieldTypeSearch();
// ...
const type = findType(field.typeClass);
```

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/modal/modal.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+  const { data } = useFetchGroups();
+
+  useEffect(() => {
+    if (data && !loaded) {
+      setState(data);
+
+      setLoaded(true);
+    }
+  }, [data, loaded]);
```
gustavs-gutmanis, line 70: I'm not sure `loaded` is needed at all. It's only used to set the state.
We can do that by using the `isFetched` from the `useFetchGroups` query.

```suggestion
  useEffect(() => {
    if (isFetched) {
      setState(data);
    }
  }, [isFetched, data]);
```

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/field-group/field-group.tsx
```diff
@@ -32,7 +33,38 @@ export const FieldGroup: React.FC<PropsWithChildren<Props>> = ({
...
 };
+
+FieldGroup.displayName = 'FieldGroup';
+
+type ManagerEditionProps<P extends FieldGroupProps> = Omit<P, 'button'> & {
+  button?: FieldGroupProps['button'];
+  editionIsPro?: boolean;
+};
+
+const withManagerEdition = <P extends FieldGroupProps>(
```
gustavs-gutmanis: I'm not a big fan of [HoC's](https://gist.github.com/heygrady/f9bf3b6dd93fe3d87ba87430fd3c20d5)
Can we just have a standard component which does this very same thing?

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/modal/model.operations.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export const useGroupOperations = (
+  initialState: Group,
+  setState: StateSetter<Group>,
+  fieldListRefs: React.MutableRefObject<FieldListRefs>
+): GroupOperations => {
+  const generateRandomColor = (): string =>
+    `#${(Math.floor(Math.random() * 0xffffff) + 0x1000000)
+      .toString(16)
+      .slice(1)}`;
```
gustavs-gutmanis: This does not have to reside inside the hook, it can be brought out of the hook's scope.

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/modal/sortable.initializer.ts
```diff
@@ -0,0 +1,76 @@
...
+    selector: keyof FieldListRefs,
+    options?: Sortable.Options
+  ): void => {
+    const createSortable = fieldListRefs.current[selector];
+    if (createSortable) Sortable.create(createSortable as HTMLElement, options);
+  };
+
+  const handleSortableInit = (): void => {
+    const putCondition = (to: Sortable, from: Sortable): boolean =>
+      from.options.handle === '.handle' ? false : true;
```
gustavs-gutmanis: This function never changes, it should be outside of the scope of this function, otherwise it will always get generated from scratch on each init.

You can put it before `initializeSortable`

packages/plugin/src/controllers/api/fields/GroupsController.php
```diff
@@ -0,0 +1,86 @@
...
+            'types' => [],
+            'groups' => [
+                'hidden' => [],
+                'grouped' => [],
+            ],
+        ];
+
+        if ($freeform->isPro()) {
+            $hiddenFieldTypes = $freeform->settings->getSettingsModel()->hiddenFieldTypes;
+            $unassignedTypes = array_diff($types, array_merge(...array_column($groups, 'types')), $hiddenFieldTypes);
```
gustavs-gutmanis: If you're doing some dark magic kung fu, then please try to make it as readable as possible.
A good thing to do is use newlines and indentation to make it more readable.
The best is using variables and good naming.

```suggestion
            $unassignedTypes = array_diff(
                $types,
                array_merge(
                    ...array_column($groups, 'types')
                ), 
                $hiddenFieldTypes
            );
```

This is pretty genius though! But takes some time to process.
So this would be a prime candidate for pulling it out into a separate, well named variable.

```php
$flattenedAssignedTypes = array_merge(...array_column($groups, 'types')), 
```


```suggestion
            $unassignedTypes = array_diff(
                $types,
                $flattenedAssignedTypes,
                $hiddenFieldTypes
            );
```

packages/plugin/src/controllers/api/fields/GroupsController.php
```diff
@@ -0,0 +1,86 @@
...
+                'grouped' => $groups,
+            ];
+
+            return $response;
+        }
+
+        $filteredGroups = array_map(function ($group) use ($types) {
+            $filteredTypes = array_filter($group->types, function ($type) use ($types) {
+                return \in_array($type, $types);
+            });
```
gustavs-gutmanis: There's a nice shorthand for writing anonymous function in php now.

```suggestion
            $filteredTypes = array_filter(
                $group->types, 
                fn ($type) => \in_array($type, $types)
            );
```

packages/plugin/src/controllers/api/fields/GroupsController.php
```diff
@@ -0,0 +1,86 @@
...
+                'uid' => $group->uid,
+                'label' => $group->label,
+                'color' => $group->color,
+                'types' => $filteredTypes,
+            ];
+        }, $groups);
+
+        $response->groups['grouped'] = array_values(array_filter($filteredGroups, function ($group) {
+            return !empty($group->types);
+        }));
```
gustavs-gutmanis: Please use newlines and indentation to make such statements readable.

```suggestion
        $response->groups['grouped'] = array_values(
            array_filter(
                $filteredGroups, 
                fn ($group) => !empty($group->types)
            )
        );
```

gustavs-gutmanis (commented):
Looking great. Some small improvements can be made here and there and then we're all good to go

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/modal/modal.tsx
thejahid: When 'Save' is clicked, the groups are saved to the database, and the modal closes. I don't want to reset the state;  It'll cause some problems.

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/base-fields.tsx
thejahid: I noticed the hook; it uses 'useFetchFieldTypes' every time and looks for the 'typeClass', Since there are many fields, I decided to make the request just once<useFetchFieldTypes>. The '<LoaderFieldGroup words={[50, 70]} items={16} />' will work nicely without needing to set it for every single item."

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/modal/modal.tsx
gustavs-gutmanis: This shouldn't impact any thing.
* The state represents the initial dataset that comes from the endpoint.
* When you edit the groups, the state changes, then you click save - a mutation is triggered.
* Once the mutation is finished, the modal is closed and the query gets invalidated and re-fetched
* At this point, the modal is already closed, so it would not change the sate. But even if it did change the state, it would change it to the way it already is, because the state is persisted to the database.

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/base-fields.tsx
gustavs-gutmanis: `useFetchFieldTypes` would always return a memoized set of data, it would not call the actual endpoint.

So in the end - it would work exactly the same as your current solution, with 0 overhead. Just that you wouldn't have to write the same code twice.

gustavs-gutmanis (approved):
Fantastic work!

PR #950, opened 2024-01-04
[SFT-642]: Blend Surveys & Polls plugin into Freeform 5

gustavs-gutmanis:
### Description
* adding survey types to FF5
* adding survey views and controllers


kjmartens (changes requested):
Looks good. 🙂  Just a couple of issues I noticed:

1. The default display of fields (before they have been touched for overriding) is not respecting the configurable default settings.
Screenshot 2024-01-04 at 12 40 30 PM

2. Encrypted field data is not being decrypted.
Screenshot 2024-01-04 at 12 40 47 PM

PR #951, opened 2024-01-05
[SFT-724] - Field handle validation shouldn't FORCE camelCase without exceptions #891

seandelaney:
Removed camel case enforcement. 
Updated existing unit test cases, adding two new unit tests.

image


gustavs-gutmanis (approved):
🆗

PR #958, opened 2024-01-10
[SFT-689]: Save button

gustavs-gutmanis:
### Description

* adding save button props to the builder
* updating code to use those props when dealing with save functionality

### Images

image


kjmartens (changes requested):
Looks great! However, I encountered 2 issues:

1. Each successive Form Save will cause there to be more than one `post` in the `method` attribute for the form when revisiting the form. For example, after the 1st save, the source code will appear as follows: `<form method="post post" data-freeform=""...`. When I remove the extra `post` string with the browser inspector, I can submit successfully, but then the next time it'll keep snowballing and you may end up with: ``<form method="post post post post post post post"...`
2. The **Save** button (and **Back** button) is ALWAYS displayed in the form layout, regardless of whether they are enabled or not.
Screenshot 2024-01-10 at 1 37 27 PM

kjmartens (approved):

PR #959, opened 2024-01-10
[SFT-813]: Moving fields around in the layout causes data loss for Text-type fields

gustavs-gutmanis:
### Description

* fixed an issue where field data would be wiped from submissions when changing fields between rows


kjmartens (approved):
Excellent! Works good now. 🙂

PR #961, opened 2024-01-11
[SFT-526] Fix page tab dropdown

aaronlinley:
The issue here is that the index for each tab is used for the `data-id` attribute in the tab HTML. If this is evaluated to anything other than a string, the JS throws an exception (visible in the console). As this tab config array was (by default) numerically indexed, the value of the `data-id` attribute (e.g. `data-id="1"`) was evaluating to a number, causing the exception to be thrown. Forcing this to be string-indexed resolves this issue.


gustavs-gutmanis:
Please run `composer run fix` to fix the linting issues, and then it's good to go.

gustavs-gutmanis (approved):
Nice 👌🏻

kjmartens (approved):
Looks great!

PR #967, opened 2024-01-12
Drag'n'Drop: Fix incorrect url if there are search params used

urbantrout:
For example, if the page uses `?token=123`


kjmartens:
Thank you for this, @urbantrout. We'll review this shortly. 🙂

gustavs-gutmanis (approved):
Beautiful solution, thank you @urbantrout

urbantrout:
You’re welcome :)

PR #968, opened 2024-01-14
feat/v5/SFT-787: Add Field associations badges

thejahid:
Field associations badges in form builder.


packages/client/src/app/pages/forms/edit/store/slices/integrations/integrations.selectors.ts
```diff
@@ -2,9 +2,57 @@ import type { RootState } from '@editor/store';
...
+            const { customerMapping = {}, addressMapping = {} } = values || {};
+            return (
+              Object.values(customerMapping).some(
+                (customerAttr) => customerAttr.value === uid
+              ) ||
+              Object.values(addressMapping).some(
+                (addressAttr) => addressAttr.value === uid
+              )
+            );
+          }
```
gustavs-gutmanis: This won't really work unfortunately, each integration can have different ways of mapping attributes.

I guess you have to extract the integration properties and check for field mapping fields.

I wonder if it's possible even.

packages/client/src/app/pages/forms/edit/store/slices/integrations/integrations.selectors.ts
```diff
@@ -2,9 +2,57 @@ import type { RootState } from '@editor/store';
...
 
+type IntegrationValues = {
+  emailField?: string;
+  optInField?: string;
+  attributeMapping?: Record<string, { value: string }>;
+  customerMapping?: Record<string, { value: string }>;
+  addressMapping?: Record<string, { value: string }>;
+};
+
+const allowedTypes = ['email-marketing', 'elements', 'payment-gateways'];
```
gustavs-gutmanis: I'm not sure we need this limitation, because CRM integrations are mapping fields the most.

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-layout/page/page.tsx
```diff
@@ -13,13 +16,28 @@ type Props = {
...
   const layout = useAppSelector((state) =>
     layoutSelectors.pageLayout(state, page?.layoutUid)
   );
+  const { isFetched: isFormRulesFetched } = useQueryFormRules(
+    Number(formId || 0)
+  );
+  const { isFetched: isFormNotificationsFetched } = useQueryFormNotifications(
+    Number(formId || 0)
+  );
+  const { isFetched: isFormIntegrationsFetched } = useQueryFormNotifications(
```
gustavs-gutmanis: This should query for integrations I assume.

gustavs-gutmanis: I would suggest putting this into `forms/edit/edit.tsx` where we're pre-fetching integrations and settings.

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-layout/field/cell/cell.tsx
gustavs-gutmanis, line 1: All of the logic added here in this file should be in a separate component, which deals with rendering the icons. There is no need for this component to know anything about the fetching logic.

So I would suggestputting the logic in the `cell-badges.tsx`

packages/client/src/app/pages/forms/edit/store/slices/integrations/integrations.selectors.ts
```diff
@@ -2,9 +2,57 @@ import type { RootState } from '@editor/store';
 
 import type { IntegrationEntry } from '.';
 
+type IntegrationValues = {
+  emailField?: string;
+  optInField?: string;
+  attributeMapping?: Record<string, { value: string }>;
+  customerMapping?: Record<string, { value: string }>;
+  addressMapping?: Record<string, { value: string }>;
+};
```
gustavs-gutmanis: These are never guaranteed to be like this, each integration can have different ways of mapping fields.

If it's not possible to infer this from the integration properties, then we either have to check all the properties for an object that might resemble a field mapping, or just skip showing this integration association badge entirely

gustavs-gutmanis (changes requested):

packages/client/src/app/pages/forms/edit/store/slices/integrations/integrations.selectors.ts
thejahid: I went through every single integration type - [email-marketing, elements, payment-gateways] and found the mapping attributes. For example, email marketing only has 2 boxes where we can add 2 fields: 1. Target Email Field, 2. Opt-in Field (optional).

gustavs-gutmanis: In Salesforce, for instance, there are 4 different field mappings.

* account
* contact
* lead
* opportunity

`packages/plugin/src/Integrations/CRM/Salesforce/Versions/SalesforceV58.php`

Others have different mappings.
We cannot manually determine which properties to check, that is not a viable solution, so it either has to be automatic, or shouldn't be there at all.

What I would do is - check the integration properties list, filter out all of the `FieldMapping` properties, and look inside the values for the current field uid.

If that's not doable, I would look at all properties which look like field mappings, and check that way.

If that can't be done - I would just scratch this feature entirely and not do it.

gustavs-gutmanis (approved):
Nice! 👌🏻

kjmartens (changes requested):
Seems to work great now, however there seems to be a bit of a performance issue.

Before the addition of this feature, it was around 3s to load the builder (which isn't too great already I guess). It's now around 4-4.5s. What's worse is that roughly 1/5 refreshes it just hangs in the loading state and never loads.

kjmartens (approved):

kjmartens:
Splendid work 👌

PR #969, opened 2024-01-14
fix/v5/SFT-826: fix t.fields.filter

thejahid:
Field search in builder is not working when searching across forms. Gives `t.fields.filter is not a function` error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Solspace Freeform Changelog
 
+## 5.0.0-beta.17 - Unreleased
+
+> [!WARNING]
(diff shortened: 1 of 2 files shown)


gustavs-gutmanis:
What does this fix change?
It looks like the change is the same in terms of what it does as before.

thejahid:
Yes but the previous one give object sometimes instead of array

gustavs-gutmanis:
Ok, so then do this:

```php
$fields = array_map([$this->fieldTransformer, 'transform'], $fields);
$fields = array_values($fields);
```

thejahid:
Yeah, that's correct, but isn't it re-indexing all the fields again? So I thought I'll do it one time.

kjmartens (approved):
Looks good

PR #974, opened 2024-01-19
[SFT-823] - Quick Form Widget - Updated to use Grid layout template

seandelaney:
`customAttributes` doesn't exist anymore. Replaced with `attributes` but even at that, sub attributes had also changed. 

Then new template styled didn't work well within the widget so force set the template to grid layout with minor tweaks.


kjmartens (approved):
👌

PR #977, opened 2024-01-22
[SFT-838] - Setting fallback success and error messages if blank in the form builder

seandelaney:
Set fallback success and error messages if left blank in the form builder


gustavs-gutmanis (approved):
Nice 👌🏻

PR #986, opened 2024-01-26
[SFT-847]: fixing muti-value field conditional values

gustavs-gutmanis:
* fixing checkboxes
* fixing multiple value select boxes
* fixing single checkboxes


kjmartens (approved):
👌

Issue #988, opened 2024-01-26
v5 update migration freeform_rules_pages exception


RyVeata:
**Describe the bug or issue you're experiencing**
Updating from 4.1.13 to version 5.0.2 fails due to a migration exception as follows:
`add primary key PRIMARY_KEY on {{%freeform_rules_pages}} (id) ...Exception: SQLSTATE[42P07]: Duplicate table: 7 ERROR:  relation "PRIMARY_KEY" already exists
The SQL being executed was: ALTER TABLE "freeform_rules_pages" ADD CONSTRAINT "PRIMARY_KEY" PRIMARY KEY ("id") (/usr/share/nginx/vendor/yiisoft/yii2/db/Schema.php:676)`.  I have experienced this error on my build both from an existing install of Freeform that I had upgraded to 4.1.13, and from a clean install version of 4.1.13.

**Steps to reproduce**

1. Install Freeform 4.1.13
2. Run `craft update/update freeform:5.0.2` on the CLI

**Expected behavior**
The update and migrations would complete successfully allowing Freeform to function on version 5.

**Craft & Plugin Info (please complete the following information):**

- Craft Version: Craft Solo 4.7.0
- Freeform Version: Freeform 4.1.13
- Freeform Edition: Freeform Pro
- Fresh Install or Upgrade: Both

**Additional context**
The freeform_rules_pages table does not exist in the DB prior to running the migration, and the table itself get successfully created in the previous line `create table {{%freeform_rules_pages}} ... done (time: 0.001s)`.  Freeform tables are as follows at the point of upgrade:

```freeform_forms_layouts
 freeform_forms_pages
 freeform_forms_rows
 freeform_forms_fields
 freeform_favorite_fields
 freeform_forms_notifications
 freeform_forms_integrations
 freeform_crm_fields
 freeform_export_profiles
 freeform_export_settings
 freeform_feeds
 freeform_fields
 freeform_integrations_queue
 freeform_lock
 freeform_mailing_list_fields
 freeform_mailing_lists
 freeform_notification_log
 freeform_notifications
 freeform_payment_gateway_fields
 freeform_payments_payments
 freeform_payments_subscription_plans
 freeform_payments_subscriptions
 freeform_saved_forms
 freeform_session_context
 freeform_spam_reason
 freeform_statuses
 freeform_submission_notes
 freeform_unfinalized_files
 freeform_webhooks
 freeform_webhooks_form_relations
 freeform_notification_templates
 freeform_forms
[...]

PR #993, opened 2024-01-29
[SFT-850] - Postgres issue with v5 update migration

seandelaney:
https://github.com/solspace/craft-freeform/issues/988

- When installing, only run `addPrimaryKey` (if using MySQL) and `addForeignKey` after install so order is correct.
- When upgrading from v4 to v5, only run `addPrimaryKey` if using MySQL. 
- Fixed issue with Predefined Options, not being passed configuration data.

**Notes**
Postgres is setting Primary Keys automatically.


gustavs-gutmanis (approved):
Very nice work 👌🏻

PR #994, opened 2024-01-30
[SFT-856]: fixing digest emails displaying 0's and not showing error logs.

gustavs-gutmanis:
* adding reference dates to digest jobs
* fixing an issue with submission count lookup


kjmartens (approved):
👌

PR #1003, opened 2024-01-31
feat/v5/SFT-828

thejahid:
### Description

Add calculation field.


packages/client/src/app/components/form-controls/control-types/calculation-box/calculation-box.tsx
```diff
@@ -0,0 +1,103 @@
...
+                  tagData
+                )}>
+                <x title='remove tag' class='tagify__tag__removeBtn'></x>
+                  <div>
+                    <span class='tagify__tag-text'><span class='sr-only-value'>field:</span>${
+                      tagData.value
+                    }${' '}</span>
+                  </div>
+                </tag>`;
+              },
```
seandelaney: Did eslint format this block of code?

Watch out for mixed single quotes and double quotes.

This looks much more cleaner and readable (but not tested):

```
tag: function (tagData) {
    return `
        <tag
            title="${tagData.value}"
            contenteditable="false"
            spellcheck="false"
            class="tagify__tag"
            ${this.getAttributes(tagData)}
        >
            <x title="remove tag" class="tagify__tag__removeBtn"></x>
            <div class="tagify__tag-text">
                <span class="sr-only-value">field:</span>${tagData.value}${' '}
            </div>
        </tag>
    `;
},
```

Also, do we need the `sr-only-value` class ?
... and `${' '}` to create a space?

thejahid: It's easier to get the value from textContent and convert it to a tag. So, I add the 'sr-only-value' & for the space - This is because I retrieve the value from textContent. If anyone forgets to add space between the operator and the tag, it will be added.

packages/client/src/app/components/form-controls/control-types/calculation-box/calculation-box.tsx
```diff
@@ -0,0 +1,107 @@
...
+        'Solspace\\Freeform\\Fields\\Implementations\\DropdownField',
+        'Solspace\\Freeform\\Fields\\Implementations\\RadiosField',
+        'Solspace\\Freeform\\Fields\\Implementations\\HiddenField',
+        'Solspace\\Freeform\\Fields\\Implementations\\Pro\\InvisibleField',
+        'Solspace\\Freeform\\Fields\\Implementations\\Pro\\OpinionScaleField',
+        'Solspace\\Freeform\\Fields\\Implementations\\Pro\\RatingField',
+        'Solspace\\Freeform\\Fields\\Implementations\\Pro\\RegexField',
+      ].includes(item.typeClass)
+    )
+    .map((item) => item.properties.handle);
```
gustavs-gutmanis: There's a [Field attribute](https://github.com/solspace/craft-freeform/blob/v5/packages/plugin/src/Attributes/Property/Input/Field.php) which has an array of `implements` defined.

Whenever I set that property on a field, I specify which type it should implement there (be it a `NumericInterface`, or `OptionsInterface`, etc).

In this case it would be a simpler case - just listing the available type classes which should be allowed here.

Then the Property would contain this information, and you could access it via `property.availableFieldTypes` or something.

Thus not hardcoding this information into react.

packages/plugin/src/Attributes/Property/Input/CalculationBox.php
```diff
@@ -0,0 +1,14 @@
...
+
+use Solspace\Freeform\Attributes\Property\Property;
+
+/**
+ * @extends Property<string>
+ */
+#[\Attribute(\Attribute::TARGET_PROPERTY)]
+class CalculationBox extends Property
+{
+    public ?string $type = 'calculationBox';
```
gustavs-gutmanis, line 13: We would add the `availableFieldTypes` array here. Check the [`Field` Attribute](https://github.com/solspace/craft-freeform/blob/v5/packages/plugin/src/Attributes/Property/Input/Field.php) for reference.

packages/plugin/src/Fields/Implementations/Pro/CalculationField.php
```diff
@@ -0,0 +1,51 @@
...
+    name: 'Calculation',
+    typeShorthand: 'calculation',
+    iconPath: __DIR__.'/../Icons/calculator.svg',
+    previewTemplatePath: __DIR__.'/../PreviewTemplates/calculation.ejs',
+)]
+class CalculationField extends TextField
+{
+    public const DEFAULT_CALCULATIONS = '';
+    protected string $instructions = '';
+    protected string $placeholder = '';
```
gustavs-gutmanis: Lets have a space between constants and variable definitions.
Also there's no need to make it public, since nothing else uses it.

```suggestion
    private const DEFAULT_CALCULATIONS = '';

    protected string $instructions = '';
    protected string $placeholder = '';
```

packages/plugin/src/Fields/Implementations/Pro/CalculationField.php
```diff
@@ -0,0 +1,51 @@
...
+
+    #[Input\CalculationBox(
+        label: 'Calculation Logic',
+        instructions: 'Type "@" or "{" to get the field for calculation.',
+    )]
+    protected string $calculations = self::DEFAULT_CALCULATIONS;
+
+    public function getType(): string
+    {
+        return self::TYPE_CALCULATION;
```
gustavs-gutmanis, line 63: Leave this as is, but check comment below.

packages/plugin/src/Fields/Implementations/Pro/CalculationField.php
```diff
@@ -0,0 +1,51 @@
...
+        return $this->calculations;
+    }
+
+    public function getInputHtml(): string
+    {
+        $attributes = $this->getAttributes()
+            ->getInput()
+            ->clone()
+            ->setIfEmpty('name', $this->getHandle())
+            ->setIfEmpty('type', $this->getType())
```
gustavs-gutmanis: This should be type `text`.

packages/plugin/src/Fields/Implementations/Pro/CalculationField.php
```diff
@@ -0,0 +1,51 @@
...
+            ->setIfEmpty('id', $this->getIdAttribute())
+            ->setIfEmpty('value', $this->getValue())
+            ->setIfEmpty('data-calculations', $this->getCalculations())
```
gustavs-gutmanis: This should be `replace` because we should not allow others to edit this particular attribute.
```suggestion
            ->replace('data-calculations', $this->getCalculations())
```

packages/scripts/src/lib/plugin/handlers/fields/calculations.ts
```diff
@@ -0,0 +1,112 @@
...
+      return true;
+    } else if (lowercasedValue === 'false') {
+      return false;
+    }
+
+    return isNaN(Number(value)) ? value : Number(value);
+  };
+
+  reload = () => {
+    const pickers = this.freeform.form.querySelectorAll('input[type="calculation"]');
```
gustavs-gutmanis: There are no valid HTML input types called `calculation`. I'm not sure this is a correct approach. The type should be `text` for the input itself.

I think you can select the pickers by the attribute.

```suggestion
    const pickers = this.freeform.form.querySelectorAll('input[data-calculations]');
```

packages/scripts/src/lib/plugin/handlers/fields/calculations.ts
```diff
@@ -0,0 +1,112 @@
...
+        if (allVariablesHaveValues) {
+          const result = this.expressionLanguage.evaluate(calculationsLogic, variables);
+
+          if (picker instanceof HTMLInputElement) {
+            picker.value = result;
+          }
+        }
+      };
+
+      Object.keys(variables).forEach((variable) => {
```
gustavs-gutmanis: This whole block needs some refactoring. It's way too complex.

Read up on [Guard Clauses](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html), to prevent such nesting of IF's.

I can notice that every conditional block uses pretty much the same variable lookup and calls the same function.

I think something can be done here to make the code repeat less and improve its readability.

You can also focus on other things and I will later see how this can be refactored and then we can discuss the refactor.

gustavs-gutmanis (commented):
Great work. Lets continue with the functionality on the backend part by writing a Validator which will validate that the calculation coming from the `POST` request is valid.

packages/plugin/src/Bundles/Fields/Validation/CalculationFieldValidation.php
```diff
@@ -0,0 +1,89 @@
...
+        $field = $event->getField();
+        if (!$field instanceof CalculationField) {
+            return;
+        }
+
+        $form = $event->getForm();
+        $value = $this->valueOrdination($field->getValue());
+        $calculations = $field->getCalculations();
+
+        $getVariablesPattern = '/field:([a-zA-Z0-9_]+)/u';
```
gustavs-gutmanis: This should be a `private const` for the class.
Otherwise you're always creating an instance of this variable for no reason.

packages/plugin/src/Bundles/Fields/Validation/CalculationFieldValidation.php
```diff
@@ -0,0 +1,89 @@
...
+        preg_match_all($getVariablesPattern, $calculations, $matches);
+        $variables = $matches[1];
+
+        $calculationsLogic = $calculations;
```
gustavs-gutmanis: There's no difference between these variables, and `$calculations` is never used afterwards.

In this case, either use `$calculations` or name the initial variable `$calculationLogic`.

packages/plugin/src/Bundles/Fields/Validation/CalculationFieldValidation.php
```diff
@@ -0,0 +1,89 @@
...
+        $variablesWithValue = [];
+
+        foreach ($variables as $variable) {
+            $calculationsLogic = str_replace("field:{$variable}", $variable, $calculationsLogic);
+
+            $getValue = $form->get($variable)->getValue();
```
gustavs-gutmanis: `$getValue` kind of implies a value from the `$_GET` global array.
Maybe it would be best to rename `$value` to `$valueOrdination` and call this one `$value` or `$fieldValue` instead.

packages/plugin/src/Bundles/Fields/Validation/CalculationFieldValidation.php
```diff
@@ -0,0 +1,89 @@
...
+            if (null !== $getValue && '' !== trim($getValue)) {
+                $variablesWithValue[$variable] = $this->valueOrdination($getValue);
+            } else {
+                $field->addError(Freeform::t('Variable "{variable}" is missing a value', ['variable' => $variable]));
+
+                return;
+            }
+        }
+
+        $calculationsLogic = str_replace(['&#8203;', ' ', '\\u200B', "\xE2\x80\x8B", "\xC2\xA0"], ' ', $calculationsLogic);
```
gustavs-gutmanis: What's happening here? What do the hex codes stand for?

packages/plugin/src/Bundles/Fields/Validation/CalculationFieldValidation.php
```diff
@@ -0,0 +1,89 @@
...
+            if ($value !== $result) {
+                $field->addError(Freeform::t('Incorrectly calculated value'));
+            }
+        } catch (SyntaxError $e) {
+            $field->addError(Freeform::t('Syntax error in calculation: {error}', ['error' => $e->getMessage()]));
+        } catch (\TypeError $e) {
+            $field->addError(Freeform::t('Type error in calculation: {error}', ['error' => $e->getMessage()]));
+        } catch (\Exception $e) {
+            $field->addError(Freeform::t($e->getMessage()));
+        }
```
gustavs-gutmanis: It's best not to expose the error messages that the server generates. It might expose sensitive information.

Better log the error using the Freeform Logger and add a generic error message `Error in calculation` or something.

packages/plugin/src/Fields/Implementations/Pro/CalculationField.php
```diff
@@ -0,0 +1,91 @@
...
+            TextField::class,
+            NumberField::class,
+            TextareaField::class,
+            DropdownField::class,
+            RadiosField::class,
+            HiddenField::class,
+            InvisibleField::class,
+            OpinionScaleField::class,
+            RatingField::class,
+            RegexField::class,
```
gustavs-gutmanis, line 47: Beautiful

packages/plugin/src/Fields/Implementations/Pro/CalculationField.php
gustavs-gutmanis: This is still not `text`

packages/plugin/src/Fields/Implementations/Pro/CalculationField.php
```diff
@@ -0,0 +1,91 @@
...
+        ]
+    )]
+    protected string $calculations = self::DEFAULT_CALCULATIONS;
+
+    public function getType(): string
+    {
+        return self::TYPE_TEXT;
```
gustavs-gutmanis: Sorry for the confusing naming here.

`getType` is the shorthand field type, in this case `calculation` is correct.

But when generating an HTML input, the `type` should be `text` because HTML does not have a native `calculation` input type.

gustavs-gutmanis (changes requested):
Good work. Left a couple of comments here and there.

packages/plugin/src/templates/submissions/edit.twig
```diff
@@ -87,6 +87,25 @@
...
+                                    value: submission[fieldHandle],
+                                    required: field.required,
+                                    errors: submission.errors(fieldHandle)
+                                } %}
+
+                                {{ forms.textField(options) }}
+                            </div>
+                        </div>
+                    {% endfor %}
+
```
gustavs-gutmanis: This seems like it renders a calculations field out of order here. What's the reasoning for that? To not render it as hidden?

There's a better way for that, you can override each specific type rendering with a template of its own.

I've added a refactor for this

gustavs-gutmanis (commented):
There's a bit of an issue with how the calculation box parses values repeatedly in the builder.

gustavs-gutmanis (dismissed):
Looks really good, lets wait for @kjmartens to approve as well

kjmartens (dismissed):
🎉

kjmartens (approved):
🎉

Issue #1014, opened 2024-02-01
HubSpot integration not working in v5


FreekVR:
**Describe the bug or issue you're experiencing**
I'm trying to sync contacts to Hubspot in v5. I'm having several issues unforunately.

It looks like the Contact sync ONLY triggers if deals and companies endpoints are also enabled and mapped to this field.

It looks like there is a bug leading to errors if the "IP Adress" is not mapped in the settings for the integration:
```
 "debugMessage": "Solspace\\Freeform\\Integrations\\CRM\\HubSpot\\BaseHubSpotIntegration::getIpField(): Return value must be of type string, null returned",
```

**Steps to reproduce**

1. Enable a hubspot integration, leaving empty the optional IP address mapping field
2. Connect it to a form and map only contacts
3. Attempt to submit the form (I used graphql mutation to do so)
4. No hubspot API calls are made
5. Map companies and deals as well, and attempt to submit again
6. Error on getIpField return type

**Expected behavior**
I should be able to sync contacts without having to also map and enable Companies and Deals AND enter some information for the ip adress in the itnegration settings.

**Craft & Plugin Info (please complete the following information):**

- Craft Version: 4
- Freeform Version: 5.0.5
- Freeform Edition: Pro
- Fresh Install or Upgrade: Upgrade

**Additional context**
There are some code quality issues in HubSPotV1.php. There are some early returns which make no sense and cause code to not be executed even though it should/could be.

```php
        if (!$this->mapDeals) {
            return;
        }

        $contactEmail = $this->getEmailFieldValue($this->contactProps);

        if (!$this->companyProps) {
            return;
        }

         // company sync happens 

        if ($this->contactProps) {
             // contacts sync happens
         }

        if (!$this->dealProps) {
            return;
        }

       // deal sync happens
```

As you can see:
- No sync will happen if deal mapping is disbled
- No sync will happen if there are no props mapped in the company sync

I would expect each code block to be executed independently of each other.

FreekVR:
As a side note, after switching to a fork with fixes from the PR, I am still getting intermittent "EXPIRED_AUTHENTICATION" errors back from Hubspot. When I repeat the request a second time it works, so it appears something isn't getting refreshed in time.

kjmartens:
Thank you very much for this @FreekVR. We are revisiting the HubSpot integration right now and will be releasing a larger update for this shortly. 🙂

kjmartens:
This should be resolved in Freeform 5.0.7 now. 🙂

PR #1016, opened 2024-02-01
Fix several hubspot sync issues

FreekVR:
### Description
See #1014 
- Fixes error when hubspot IP field is empty
- Fixes sync errors not reported if occuring during contact creation rather than update
- Fixes contacts not syncing without company/deal also enabled


kjmartens:
Thank you very much for this @FreekVR. We are revisiting the HubSpot integration right now and will be releasing a larger update for this shortly. 🙂

gustavs-gutmanis:
Thanks a lot for the PR @FreekVR 
The flow we used was quite outdated and we opted to refactor the whole thing instead to use their `v3` REST API as well as switch from Public Apps to Private Apps which made more sense in this scenario, because of how convulated their Public App making process is.

Here's a link to the refactor PR if you're interested https://github.com/solspace/craft-freeform/pull/1021

I will be closing your PR, and I'm very sorry that we're not using yours, but the refactor was necessary.

Issue #1009, opened 2024-01-31
Attempting to query "maxLength" for a field via GraphQL always results in an error


FreekVR:
**Describe the bug or issue you're experiencing**

**Steps to reproduce**

1. Run the following GraphQL query
```
query MyQuery {
  freeform {
    form(id: 7) {
      fields {
        ... on FreeformField_Text {
          maxLength
        }
      }
    }
  }
}
```
2. Error shown: 
```"
debugMessage": "Can't get a way to read the property \"maxLength\" in class \"Solspace\\Freeform\\Fields\\Implementations\\TextField\".",
      "message": "Internal server error",
```

**Expected behavior**
It should return the configured maximum length, or not be available to the GQL interface if unsupported.

**Craft & Plugin Info (please complete the following information):**

- Craft Version: 4
- Freeform Version: 5.0.5
- Freeform Edition: Pro
- Fresh Install or Upgrade: Upgrade

**Additional context**

kjmartens:
This should be resolved in Freeform 5.0.7 now. 🙂

Issue #1007, opened 2024-01-31
GraphQL caches not reset after editing field handle(s)


FreekVR:
**Describe the bug or issue you're experiencing**
After updating a form with a new field handle this change is not immediately reflected in graphql queries fetching the form.

**Steps to reproduce**

1. Make sure caching incl. graphql query caching is enabled on the environment
2. Open an existing freeform form
3. Change the field handle, in my case, I was trying to workaround #1005 by changing a field handle from 'course-id' to 'courseId'
4. Run a `form(id: x)` query to fetch the field handles `pages { rows { fields { handle } } }`

**Expected behavior**
The updated field handle is returned

**Craft & Plugin Info (please complete the following information):**

- Craft Version: 4.x
- Freeform Version: 5..0.5
- Freeform Edition: Pro
- Fresh Install or Upgrade: Upgrade

**Additional context**
Manually emptying the graphql cache via "utilities" fixed the issue.

kjmartens:
This should be resolved in Freeform 5.0.7 now. 🙂

Issue #1005, opened 2024-01-31
Freeform v5 defines GraphQL field names with hypens, which is not supported in gql


FreekVR:
**Describe the bug or issue you're experiencing**
I get errors like these after upgrading an old Craft 3 freeform install and running GraphQL mutations or trying to access the schema definition (introspection)
```
{
  "errors": [
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"course-id\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"course-id\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"course-id\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"course-id\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"course-id\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"course-id\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"course-id\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"course-id\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"save_safe-6-essentialsCOM_Submission\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"save_safe-6-essentialsFR_Submission\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"save_safe-6-essentialsNL_Submission\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"save_safe-6-essentialsIN_Submission\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"save_safe-6-essentialsDE_Submission\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"Course-id\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"course-id\" does not.",
    "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but \"course-id\" does not.",
[...]

Issue #1004, opened 2024-01-31
Missing GraphQL fields/interface for (page) submit buttons?


FreekVR:
After upgrading to Freeform v5, there doesn't seem to be any way to get the configuration for the (page) submit button, specifically the label and layout configs:

![image](https://github.com/solspace/craft-freeform/assets/417416/1ceb1f63-0d0d-4655-a4b1-057d3379c515)

**Steps to reproduce**

1. Create an (empty) form with the default submit in place
2. Attempt to fetch all the information via GraphQL ("full schema") graphiql

**Expected behavior**
There should be a field, or metadata in the page field, to get the information shown in the screenshot above. There doesn't appear to be any type definition or field for this?

**Craft & Plugin Info (please complete the following information):**

- Craft Version: 4.7.1
- Freeform Version:  5.0.5
- Freeform Edition: Pro
- Fresh Install or Upgrade: Upgrade (Freeform 3 -> 4 -> 5)

**Additional context**

kjmartens:
This should be resolved in Freeform 5.0.7 now. 🙂

PR #1017, opened 2024-02-01
fix(SFT-863): GraphQL

seandelaney:
- [Convert Form and Field handles with dashes to camel case](https://github.com/solspace/craft-freeform/issues/1005) 
- [Reset GraphQL cache after making form updates](https://github.com/solspace/craft-freeform/issues/1007)
- [Return a fields value/default value in GraphQL queries](https://github.com/solspace/craft-freeform/pull/1017)
- [Removed max length from text and textarea field types as not applicable](https://github.com/solspace/craft-freeform/issues/1009)
- [Missing GraphQL fields/interface for (page) submit buttons](https://github.com/solspace/craft-freeform/issues/1004)


kjmartens (approved):
Removing hyphen-conversion migrations because they don't factor in submissions tables and they can break legacy templates.

PR #1021, opened 2024-02-02
[SFT-864][#1014]: HubSpot refactor

### LINKED ISSUE #1014: HubSpot integration not working in v5

gustavs-gutmanis:
* refactored HubSpot to use API v3
* refactored HubSpot to use Private Apps and tokens
* refactored the process of mapping to and between HS Records
* removed HubSpot v1
* added version number to integration dropdown
* removed filtering of empty values from mapped value stack
* adding `stack` to register authorized client event, letting clients add middleware to the guzzle clients


kjmartens (approved):
Looks good

PR #1025, opened 2024-02-06
Various bug fixes

gustavs-gutmanis:
## Rules
* fixed non-multiple-value rules being broken
* added validation for `is empty`, `is not empty`, `is one of` and `is not one of` rules
* removed `is one of` and `is not one of` from the currently accessible options, as it will be implemented later


kjmartens (dismissed):
Currently I can only get conditional rules to work on fields where the rule is based on a Radio, Checkboxes or single Checkbox field type. Other text inputs and dropdowns don't appear to work at all.

PR #1045, opened 2024-02-09
[SFT-884][fix]: `attributes` overrides in templates and subsequent requests

gustavs-gutmanis:
### Related Ticket Number

SFT-884

### Description

There was an issue with template overrides of field attributes when the form moves forward in pages. They weren't persisted in the order they were executed initially, so some attributes could go missing.

This is alleviated now by persisting each attribute override list in a stack, and re-executing them in the correct order upon context retrieval.

- implementing storage of attribute override stack
- iterating through the override stack in the same order on consequent form loads

### DIFF
--- a/packages/plugin/src/Attributes/Property/Implementations/Options/OptionCollection.php
+++ b/packages/plugin/src/Attributes/Property/Implementations/Options/OptionCollection.php
@@ -137,7 +137,7 @@ class OptionCollection implements CustomNormalizerInterface, \IteratorAggregate,
         throw new \BadMethodCallException('OptionCollection is read-only');
     }
 
-    public function offsetUnset(mixed $offset)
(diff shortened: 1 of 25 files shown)


kjmartens (changes requested):
It seems that overrides (at any level - inside the formatting template, the calling template, etc) do not work for `buttons` -> `attributes`. The `submitLabel`, etc work, but the attributes do NOT. They seem completely dead:

``` twig
    buttons: {
        attributes: {
            container: { class: "freeform-button-container" },
            column: { class: "freeform-button-column" },
            buttonWrapper: { class: "freeform-button-wrapper" },
            submit: { class: "freeform-button-submit" },
            back: { class: "freeform-button-back" },
            save: { class: "freeform-button-save" },
        },
    },
```

kjmartens (approved):

Issue #1053, opened 2024-02-15
Fields are filled with values set via get parameters even with the "Fill Form Values from the GET Query String" is disabled


josdigital:
Fields in a form embedded via the function craft.freeform.form(handle).render() can be filled with GET query parameters even if the settings field "Fill Form Values from the GET Query String?" in Form Behavior is disabled.

**Steps to reproduce**

1. Create a form with a text field or a textarea field
2. Embed the form in a entry
3. Open the article on the frontend and add the parameter corresponding to the name attribute of the field. Eg: "/my-article?fieldName=injected-content

**Expected behavior**
With the settings "Form Behavior > Fill Form Values from the GET Query String?" set to off (disabled) it should not be possible to fill the fields with a GET query string.

**Craft & Plugin Info:**

- Craft Version: 3.9.5 
- Freeform Version: 3.13.32
- Freeform Edition: Pro
- Fresh Install or Upgrade: Upgrade

kjmartens:
Thanks for reporting this @josdigital. We'll check into it shortly. 🙂

kjmartens:
This is now released as an update in Freeform 3.13.35 as well as v4 and v5. 🙂

PR #1084, opened 2024-02-21
fix(SFT-911) Fill Form Values from the GET Query String

seandelaney:
Fixes https://github.com/solspace/craft-freeform/issues/1053


kjmartens (approved):
👌

PR #1099, opened 2024-02-29
[SFT-171]: Import/Export entire forms & submissions utility

gustavs-gutmanis:
### Related Ticket Number

SFT-171

### Description

There are other form plugins or methods of storing submission data, which need to be imported into Freeform.
This PR is the preliminary work of adding interfaces and methods for generating exports and then importing from the exported data.

- add `FreeformImporter` capable of handling generated DTO transfer into freeform
- add `ExporterInterface` defining  the way an exporter has to work and which data to provide
- add `ExpressFormsImporter` capable of importing forms, notifications and submissions from ExpressForms into Freeform


seandelaney:
When I click on Import from Express Forms, I get 500 internal server error in the console:

image
image

I do not have Express Forms installed, which is why its happening but we should catch this error.

seandelaney:
I then installed Express Forms, setup a form. When back to Freeform > Import Express Forms and got this:

image
image

kjmartens (approved):

PR #1100, opened 2024-03-01
[SFT-935]: Stripe Payment Element briefly shows error after submit and before success - possibly due to having 2 Stripe fields in same form?

gustavs-gutmanis:
### Related Ticket Number

SFT-935

### Description

When adding several payments to a single form, hiding/showing them via rules, a recent change had made it so that both get loaded at the same time and were being submitted at the same time.
One going through, the other throwing an error. This caused the form to briefly display an error from one payment, while the other was successful.

- fixed hidden field initialization to pre-hide fields on render
- updated the stripe integration to have better checks for nested hidden fields


kjmartens (changes requested):
Responded in Jira

kjmartens (approved):
Passing for now, but there is still more to this fix needed I think.

PR #1106, opened 2024-03-04
feat(SFT-890) Craft 5 Compatibility

seandelaney:
Migrated Craft3-Common functionality across and removed the old library as a dependency.

Mainly adding missing type declarations, minor tweaks to keep PHP8.2 happy.

A few Craft 5 specific functions added.

A few Craft 4 backwards compatibility changes.

I guess the main change to review is the Form controller where we handle POST and PUT requests. In Craft 4 we were returning the form as an array but not actually doing anything with the response data. Symfony Serialiser 6.0 seemed to be happy with this but in Craft 5, Symfony Serialiser 6.4 was not too happy that we were throwing in the kitchen sink, so I removed the response data altogether until we figure out what we want to do here.


packages/plugin/src/controllers/api/FormsController.php
```diff
@@ -110,7 +110,7 @@ protected function post(int|string $id = null): null|array|object
 
         $this->response->statusCode = $event->getStatus() ?? 201;
 
-        return $event->getResponseData();
+        return null;
```
gustavs-gutmanis, line 113: I'm assuming this is safe? We aren't using response data anywhere?

packages/plugin/src/migrations/m190529_135307_AddWebhookTables.php
```diff
@@ -46,21 +47,21 @@ public function safeUp()
...
+            ForeignKey::CASCADE
         );
         $this->addForeignKey(
             null,
             '{{%freeform_webhooks_form_relations}}',
             'formId',
             '{{%freeform_forms}}',
             'id',
-            'CASCADE'
+            ForeignKey::CASCADE
```
gustavs-gutmanis, line 58: Thanks for doing all these, lets keep them, but in general, inside migrations it is best to use as little code from outside of the migration as possible, because, in theory, if you would want to roll back migrations to a point where `ForeignKey` class wasn't even present, you would not be able to. 

But in Craft it doesn't really apply, I guess.

gustavs-gutmanis (dismissed):
Beautiful work

kjmartens (approved):

PR #1107, opened 2024-03-05
[SFT-931]: Adding empty options support

gustavs-gutmanis:
### Related Ticket Number

SFT-931

### Description

Generated options fields were lacking the ability to add an empty option.

- adding `Empty Option` support for all options fields

### Images
image


seandelaney (dismissed):
Looks good. Works fine. Took me a minute to understand it though. Then realised it was the label you were setting, not the value.

kjmartens (approved):
Nice work. However, it should only apply to Dropdowns. Made a task in Jira though. 🙂

PR #1143, opened 2024-03-15
Addressing various bugs

gustavs-gutmanis:
### Related Ticket Number

SFT-970
SFT-964

### Description

- fixed an issue which caused the element editing to break, when another element integration was preceding the editable one.
- fixed user group being a single choice, rather than a multiple one
- fixed "enabled" mapping on craft elements


kjmartens (changes requested):
I noticed a few issues:

- [x] The migration seems to have wiped out the existing value for the user group to assign (db now shows `"userGroupIds":[]`).
- [x] Setting up the integration to assign to multiple groups seems to only apply the user to one (the first one that I can see).
- [x] When editing, the user's data (which is mapped inside the integration of the form) does not populate the corresponding form fields. I am using the following code:
``` twig
{{ freeform.form("userEdit").render({
    elementId: currentUser.id
}) }}
```
- [x] Not sure if the password is being corrupted or something along the way, but when I try to log in as that user after registration, the password always says invalid. I can't get in until I reset the password through Craft's reset password utility.

kjmartens (approved):

PR #1147, opened 2024-03-15
Send nocache headers when rendering a CSRF input

timkelty:
Hint to browsers and caching mechanisms that a cached response should not be used.
See https://github.com/solspace/craft-freeform/pull/1146


gustavs-gutmanis (approved):
Thank you!

Issue #1111, opened 2024-03-05
Table already exists error during migration


watchwood:
I'm trying to migrate our site from Craft 3 to Craft 4, and when I run the database migrations I get the following error:

Database Exception: SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'freeform_submissions_farm_renting_form_4' already exists
The SQL being executed was: CREATE TABLE freeform_submissions_farm_renting_form_4 (
id int(11) NOT NULL,
contactname_1 text,
email_3 text,
phonenumber_6 text,
type_of_farming_youre_interested_in_15 text,
housing_requirements_20 text,
amount_of_land_required_16 text,
outbuildings_required_19 text,
field_of_study_25 text,
select_valueasa_tenant_14 text
) ENGINE = InnoDb DEFAULT CHARACTER SET = utf8

Migration: Solspace\Freeform\migrations\m220330_111857_SplitSubmissionsTable

Output:

> create table {{%freeform_submissions_farm_renting_form_4}} ...Exception: SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'freeform_submissions_farm_renting_form_4' already exists
The SQL being executed was: CREATE TABLE freeform_submissions_farm_renting_form_4 (
id int(11) NOT NULL,
contactname_1 text,
email_3 text,
phonenumber_6 text,
type_of_farming_youre_interested_in_15 text,
housing_requirements_20 text,
amount_of_land_required_16 text,
outbuildings_required_19 text,
field_of_study_25 text,
select_valueasa_tenant_14 text
) ENGINE = InnoDb DEFAULT CHARACTER SET = utf8 (/home/vagrant/code/vendor/yiisoft/yii2/db/Schema.php:676)
#0 /home/vagrant/code/vendor/yiisoft/yii2/db/Command.php(1307): yii\db\Schema->convertException()
#1 /home/vagrant/code/vendor/yiisoft/yii2/db/Command.php(1102): yii\db\Command->internalExecute()
#2 /home/vagrant/code/vendor/yiisoft/yii2/db/Migration.php(322): yii\db\Command->execute()
#3 /home/vagrant/code/vendor/solspace/craft-freeform/packages/plugin/src/migrations/m220330_111857_SplitSubmissionsTable.php(118): yii\db\Migration->createTable()
[...]

PR #1160, opened 2024-03-21
Table already exists error during migration

DavidStrada:
Fixes an issue when running an upgrade from C3 to C4 `(Database Exception: SQLSTATE[42S01]: Base table or view already exists: ).`

Checks if the table already exist otherwise creates the table based off `\Craft::$app->db->schema->getTableSchema($tableName);`

[See for ref](https://github.com/solspace/craft-freeform/issues/1111)

### DIFF
--- a/packages/plugin/src/migrations/m220330_111857_SplitSubmissionsTable.php
+++ b/packages/plugin/src/migrations/m220330_111857_SplitSubmissionsTable.php
@@ -5,6 +5,7 @@ namespace Solspace\Freeform\migrations;
 use craft\db\Migration;
 use craft\db\Query;
 use craft\helpers\StringHelper;
+use Solspace\Freeform\Library\Migrations\ForeignKey;
 
 class m220330_111857_SplitSubmissionsTable extends Migration
 {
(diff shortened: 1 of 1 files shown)


packages/plugin/src/migrations/m220330_111857_SplitSubmissionsTable.php
```diff
@@ -116,15 +118,21 @@ private function createFormTable(int $id, string $formHandle, array $fieldMap):
...
+            }
+
+            $this->addForeignKey(null, $tableName, 'id', '{{%freeform_submissions}}', 'id', ForeignKey::CASCADE);
 
-        $this->addForeignKey(null, $tableName, 'id', '{{%freeform_submissions}}', 'id', ForeignKey::CASCADE);
+            return $tableName;
+        }
 
-        return $tableName;
+        return '';
```
gustavs-gutmanis, line 135: Lets return `null` here and change the return type of the method from `: string` to `: ?string`

gustavs-gutmanis (changes requested):
Looking great, just one comment and we're good to go

gustavs-gutmanis (approved):

gustavs-gutmanis:
I'll handle the change on my end.

PR #1172, opened 2024-03-29
[SFT-998]: The Submit button prematurely becomes active again before Stripe payment form is fully processed

gustavs-gutmanis:
### Related Ticket Number

SFT-998

### Description

Due to the asynchronous nature of stripe and recaptcha callbacks, the scripts were in a race condition to do their work and proceed with the form submit. This has been refactored to support a synchronous way of executing async requests in an ordered succession, so that they no longer have race conditions.

- refactored form submit event to allow for async callbacks
- executing the async callbacks synchronously, locking the thread


kjmartens (changes requested):
Noted 3 issues in Jira

kjmartens (changes requested):
The previous issues seem to be resolved now, but I am experiencing another very specific issue:

1. When using Stripe in your form on the same page as other Freeform fields, only the first submit will be stored to the database. Here's how to duplicate:
   - Create a form with Field X, Y and Z, set them all to be required (not fully necessary, but will make the duplication process easier).
   - Add a Stripe Payment field to that page.
   - Visit the form on the front end and enter the following:
      - X: `Hello`
      - Y: `World`
      - Z: _empty_
      - Stripe: _empty_
   - Submit the form. It'll return you back to the form with errors as expected.
   - Proceed to completing the form with the following:
      - X: `Hello`
      - Y: `Shenanigans` (noticed I changed this for dramatic effect) 😄 
      - Z: `Zesty`
      - Stripe: _fill out correctly_
   - Submit the form. It'll process completely and appear to have worked nicely. However, the second round of Freeform field data won't be stored. We would expect to see:
      - X: `Hello`
      - Y: `Shenanigans`
      - Z: `Zesty`
   - But instead, we see:
      - X: `Hello`
      - Y: `World` (original text)
      - Z: _empty_ (empty even though we updated it to be `Zesty`)
   - This only seems to happen when there's a Stripe field in the form AND when it's on the same page as the other fields. When I have the Stripe Payment field on a second page, the first page data seems to store correctly, though it does bring me to another issue...
2. When I place the Stripe field on a second page, visiting the second page locks the Submit buttons down "forever". That is on one existing form that had it there already. On another form, once I move or even recreate the Stripe field on a 2nd page, it never loads in the front end (get: `Could not load payment element.`) 🤷‍♂️ 
3. `console.log`'s seem to have been left behind for debugging Submit buttons/processing.

kjmartens (approved):

PR #1175, opened 2024-03-29
feat(SFT-555): Dotdigital Email Marketing Integration

seandelaney:


packages/plugin/src/Integrations/EmailMarketing/Dotdigital/BaseDotdigitalIntegration.php
```diff
@@ -0,0 +1,181 @@
...
+
+    #[Flag(self::FLAG_ENCRYPTED)]
+    #[Flag(self::FLAG_GLOBAL_PROPERTY)]
+    #[Validators\Required]
+    #[Input\Text(
+        label: 'API URL',
+        instructions: 'Enter your API specific URL (e.g. "https://{region}-api.dotdigital.com").',
+        order: 1,
+    )]
+    protected string $apiUrl = '';
```
gustavs-gutmanis: I don't think we need this as an editable property that the user fills out.
DotDigital provides an endpoint for finding out which region you belong to [here](https://developer.dotdigital.com/docs/api-end-points#:~:text=Region%20specific%20end%20points&text=If%20you%20don't%20know,the%20top%20of%20the%20page.).

You can listen for the `IntegrationsService::EVENT_BEFORE_SAVE` event to make a call to the API endpoint, find out the API url, and then fill in the variable from there. You will need a setter method exposed for it though.
Also you would need to use the Flag for making it a hidden property.

You can glean a similar way of doing this in MailChimp integration, though it uses a different event because of the OAuth 2.0 flow

packages/plugin/src/Integrations/EmailMarketing/Dotdigital/BaseDotdigitalIntegration.php
```diff
@@ -0,0 +1,181 @@
...
+    #[Flag(self::FLAG_GLOBAL_PROPERTY)]
+    #[Validators\Required]
+    #[Input\Text(
+        label: 'API User Password',
+        instructions: 'Enter your Dotdigital API user password',
+        order: 3,
+    )]
+    protected string $apiUserPassword = '';
+
+    #[Flag(self::FLAG_GLOBAL_PROPERTY)]
```
gustavs-gutmanis: It might be worth letting people set this on a per-form basis (just removing this flag)

packages/plugin/src/Integrations/EmailMarketing/Dotdigital/BaseDotdigitalIntegration.php
```diff
@@ -0,0 +1,181 @@
...
+            '' => '',
+            'Unknown' => 'Unknown',
+            'Single' => 'Single',
+            'Double' => 'Double',
+            'VerifiedDouble' => 'VerifiedDouble',
+        ],
+    )]
+    protected string $optInType = '';
+
+    #[Flag(self::FLAG_GLOBAL_PROPERTY)]
```
gustavs-gutmanis: Same here I guess

packages/plugin/src/Integrations/EmailMarketing/Dotdigital/Versions/DotdigitalV2.php
```diff
@@ -0,0 +1,106 @@
...
+            return;
+        }
+
+        $email = strtolower($email);
+
+        $contactData = [];
+        $contactData['email'] = $email;
+        $contactData['optInType'] = $this->getOptInType();
+        $contactData['emailType'] = $this->getEmailType();
+        $contactData['dataFields'] = [];
```
gustavs-gutmanis:
```suggestion
        $contactData = [
            'email' => $email,
            'optInType' => $this->getOptInType(),
            'emailType' => $this->getEmailType(),
            'dataFields' => [],
        ];
```

Unless it's a special case, we can usually initialize arrays with this notion, as it is a bit cleaner and easier to read.

gustavs-gutmanis (dismissed):
Nice work.

I suggested an improvement which would make clients lives easier, having less data to enter in our forms.

Take a look at it and see if you want to implement it.

kjmartens (approved):
I made G's suggested changes except the API URL one, but I made a separate task for that to be addressed next week instead. 🙂

PR #1180, opened 2024-04-02
[SFT-1011]: Saleforce Checkbox field mapping

gustavs-gutmanis:
* added a transformer for mapping Salesforce `boolean` fields


kjmartens (approved):
Very nice

PR #1185, opened 2024-04-03
[SFT-933]: Cleanup up dirty values for Integration Attributes Mappings and Field Mappings

gustavs-gutmanis:
* running a procedure for removing any mappings that are outdated and not present in a give field map


kjmartens (changes requested):
It seems like the switching / data issue is resolved, but now the **Field Mapping** table always forgets which fields each are mapped to. While they are stored in the database, the UI doesn't show the selected field. It doesn't seem to matter if I refresh the page, click elsewhere and come back, etc... just never shows. If I resave the form with the field mapping not loading, the data is then removed from the database.

kjmartens (approved):
Remaining issues noted in `SFT-1038`

PR #1192, opened 2024-04-09
[SFT-1033]: recaptcha & hcaptcha lazy loading

gustavs-gutmanis:
* improved lazy loading of reCaptcha and hCaptcha scripts
* refactored the scripts to be able to work without containing any custom data, reading it from the element instead


kjmartens (changes requested):
When loading 2 forms (same form twice, or 2 diff forms) on the same page with rC v2 Checkbox, neither of them display the rC checkbox field and I get errors in the console:

``` twig
<html>
<head>
</head>
<body>

{{ freeform.form("survey2024").render }}

<hr />

{{ freeform.form("survey2024").render }}

</body>
</html>
```

Screenshot 2024-04-10 at 1 06 45 AM

It only appears to be an issue with rC v2 Checkbox.

kjmartens (approved):

PR #1194, opened 2024-04-10
[SFT-1038]: Element integration field mapping

gustavs-gutmanis:
* fixing element field mapping issue


kjmartens (approved):
Looks good now

PR #1226, opened 2024-04-23
fix(SFT-1066): Submit buttons and HTML fields in GraphQL (FF4)

seandelaney:
Resolves https://github.com/solspace/craft-freeform/discussions/1201

### DIFF
--- a/packages/plugin/src/Bundles/GraphQL/Types/Generators/FieldGenerator.php
+++ b/packages/plugin/src/Bundles/GraphQL/Types/Generators/FieldGenerator.php
@@ -647,7 +647,7 @@ class FieldGenerator extends AbstractGenerator
                 } else {
                     $value = $source->getValue();
 
-                    if (empty($value) && method_exists($source,'getDefaultValue')) {
+                    if (empty($value) && \method_exists($source,'getDefaultValue')) {
                         $value = $source->getDefaultValue();
                     }
                 }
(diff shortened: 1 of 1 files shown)


packages/plugin/src/Bundles/GraphQL/Types/Generators/FieldGenerator.php
```diff
@@ -614,12 +628,32 @@ private static function getFieldValueDefinitions(string $typeName): array
...
-            'description' => "Field's default value",
+            'description' => "Field's value",
+            'resolve' => function ($source) use ($isHtml) {
+                if ($isHtml) {
+                    $value = $source->getContent();
+                } else {
+                    $value = $source->getValue();
+
+                    if (empty($value)) {
+                        $value = $source->getDefaultValue();
```
gustavs-gutmanis, line 651: does it have a `defaultValue`?

gustavs-gutmanis (dismissed):

packages/plugin/src/Bundles/GraphQL/Types/Generators/FieldGenerator.php
seandelaney: Good catch. In FF5 they do, but since this is FF4, I've updated the code to check if `getDefaultValue` exists first. Some field types have it defined but some don't.

gustavs-gutmanis (approved):

kjmartens (approved):

PR #1228, opened 2024-04-23
fix(SFT-1055): The 'per event' ALL DAY field mapping for Freeform submissions to Calendar (FF3)

seandelaney:


packages/plugin/src/Library/Connections/CalendarEvents.php
```diff
@@ -100,7 +100,7 @@ protected function buildElement(array $transformers, ElementInterface $element =
         $currentSiteId = \Craft::$app->sites->currentSite->id;
 
         $event->siteId = $currentSiteId;
-        $event->allDay = $this->isAllDay();
+        $event->allDay = empty($event->allDay) ? $this->isAllDay() : true;
```
gustavs-gutmanis, line 103: if `$event->allDay` is set to `false` or `null`, it will use `$this->isallDay()`. 
Otherwise it will set it to `true`

Is that expected behavior?

seandelaney: Yes.

gustavs-gutmanis (approved):

PR #1248, opened 2024-04-30
[SFT-1074]: Hidden Group Field refactor to use db tables

thejahid:
### Related Ticket Number

SFT-1074

### Description

Hidden Field Type group was saving data to the Freeform Settings model, and this caused issues for people using `allowAdminChanges => false`, who then were not able to persist the changes for field groups.

- refactoring hidden field type groups to use the database table instead


packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/base-fields.tsx
```diff
@@ -48,16 +48,19 @@ export const BaseFields: React.FC = () => {
...
-        <GroupWrapper key={group.uid} color={group.color}>
-          {group.types.length > 0 && (
-            <>
-              {group.label && <GroupName>{group.label}</GroupName>}
-              <List>{renderFieldItems(group.types)}</List>
-            </>
-          )}
-        </GroupWrapper>
-      ))}
+      {data.groups.grouped?.map(
```
gustavs-gutmanis: You should not have changed the front-end at all, all of this logic needs to be in the backend.
You previously returned `hidden` and `groups` in the response, this can still remain exactly like that, just pick the `hidden` fields out of the array in the backend.

gustavs-gutmanis (changes requested):

packages/client/src/app/pages/forms/edit/builder/tabs/layout/field-list/implementations/base-fields/base-fields.tsx
thejahid: of course

packages/plugin/src/controllers/api/fields/GroupsController.php
```diff
@@ -24,6 +24,13 @@ protected function post(null|int|string $id = null): null|array|object
         $hiddenTypes = $this->request->getBodyParam('hidden', []);
         FieldTypeGroupRecord::deleteAll();
 
+        if (!empty($hiddenTypes)) {
+            $groupRecord = new FieldTypeGroupRecord();
+            $groupRecord->uid = 'hidden';
+            $groupRecord->types = json_encode($hiddenTypes);
+            $groupRecord->save();
+        }
```
gustavs-gutmanis: Hmm, what if I have hidden groups set up already? Would this not trigger an error, saying that a record like this already exists?

Also I think using `uid` is not very safe, something might rely on all db rows using `UUID v4`. 

How about using a record with handle `freeform_hidden`?
And before addinga new one, try to fetch one from the db by `FieldTypeGroupRecord::findOne(['handle' => 'freeform_hidden']);`

gustavs-gutmanis (changes requested):

gustavs-gutmanis (dismissed):

gustavs-gutmanis (approved):

kjmartens (approved):

PR #1250, opened 2024-04-30
fix(SFT-1081): Set integration model properties after EVENT_BEFORE_SAVE

seandelaney:
Sets Model data after `EVENT_BEFORE_SAVE` to give any event listeners a chance to set values first.

Resolves https://solspace.atlassian.net/browse/SFT-1081


gustavs-gutmanis (approved):
Gorgeous

gustavs-gutmanis (changes requested):
Come to think of it, I'm not sure this is the best approach

gustavs-gutmanis (approved):
Nope, totally the best change ever.

kjmartens (approved):

Issue #1253, opened 2024-05-01
Solspace\Freeform\Services\MailerService::processRecipients(): Argument #1 ($recipients) must be of type Solspace\Freeform\Notifications\Components\Recipients\RecipientCollection, array given


jcdarwin:
**Describe the bug or issue you're experiencing**

I've set up an export notification that has two email addresses entered in the `Emails to Notify` field, each on their own line.

However, when the export attempts to run, we receive the following error in Sentry (and no export is produced):

```
TypeError
Solspace\Freeform\Services\MailerService::processRecipients(): Argument #1 ($recipients) must be of type Solspace\Freeform\Notifications\Components\Recipients\RecipientCollection, array given, called in /var/www/chinacapable/releases/20240426041212/chinacapable/craft/vendor/solspace/craft-freeform/packages/plugin/src/Bundles/Notifications/Export/ExportNotifications.php on line 88
```

**Steps to reproduce**

1. Set up an export notification that has two email addresses entered in the `Emails to Notify` field, each on their own line.

2. Wait for the export to process -- you should see the error mentioned above  in the logs

**Expected behavior**

The export file should be produced successfully.

**Craft & Plugin Info (please complete the following information):**

- Craft Version: 4.8.10
- Freeform Version: 5.1.15
- Freeform Edition: Pro
- Fresh Install or Upgrade: Fresh Install

**Additional context**

Looking at the code, `processRecipients`  is indeed typed to accept `$recipients` as an `RecipientCollection`:

```
    public function processRecipients(RecipientCollection $recipients): array
    {
        if (version_compare(\Craft::$app->getVersion(), '3.5', '>=')) {
            $testToEmailAddress = \Craft::$app->getConfig()->getGeneral()->getTestToEmailAddress();
            if (!empty($testToEmailAddress)) {
                return $testToEmailAddress;
            }
        }

        return $recipients->emailsToArray();
    }
```

However, where it is called (`craft/vendor/solspace/craft-freeform/packages/plugin/src/Bundles/Notifications/Export/ExportNotifications.php`) it simply decodes the string:

```
$message->setTo($mailer->processRecipients(json_decode($notification->recipients)));
```

Presumably this should instead be something like:

```
$recipients = new RecipientCollection(json_decode($notification->recipients));
$message->setTo($mailer->processRecipients($recipients));
```

gustavs-gutmanis:
Thanks for reporting this @jcdarwin, this has slipped through the cracks during the FF5 migration.

[...]

PR #1254, opened 2024-05-01
[#1253]: Export profiles breaking on destructuring recipients

gustavs-gutmanis:
### Related Ticket Number

#1253

### Description

When setting up export notifications and specifying multiple emails, the notification crashes when building the message, because it expects an instance of `RecipientCollection`

- building a `RecipientCollection` from the reipient email array and then passing it to the message.


kjmartens (approved):
Was able to duplicate the issue by attempting to delete Craft cache. Confirmed this issue is no longer happening after this fix.

PR #1273, opened 2024-05-09
feat(SFT-977): Added validation settings to the Password field type to force a stronger password

seandelaney:
### Related Ticket Number

[SFT-977](https://solspace.atlassian.net/browse/SFT-977)

[SFT-977]: https://solspace.atlassian.net/browse/SFT-977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/plugin/src/Bundles/Fields/Validation/CharacterVariabilityValidation.php
+++ b/packages/plugin/src/Bundles/Fields/Validation/CharacterVariabilityValidation.php
@@ -4,13 +4,15 @@ namespace Solspace\Freeform\Bundles\Fields\Validation;
 
 use Solspace\Freeform\Events\Fields\ValidateEvent;
 use Solspace\Freeform\Fields\FieldInterface;
-use Solspace\Freeform\Fields\Interfaces\MaxLengthInterface;
+use Solspace\Freeform\Fields\Interfaces\CharacterVariabilityInterface;
 use Solspace\Freeform\Freeform;
 use Solspace\Freeform\Library\Bundles\FeatureBundle;
 use yii\base\Event;
 
-class MaxLengthValidation extends FeatureBundle
+class CharacterVariabilityValidation extends FeatureBundle
 {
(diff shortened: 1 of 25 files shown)


packages/plugin/src/Bundles/Fields/Validation/CharacterVariabilityValidation.php
```diff
@@ -0,0 +1,47 @@
...
+use Solspace\Freeform\Events\Fields\ValidateEvent;
+use Solspace\Freeform\Fields\FieldInterface;
+use Solspace\Freeform\Fields\Interfaces\CharacterVariabilityInterface;
+use Solspace\Freeform\Freeform;
+use Solspace\Freeform\Library\Bundles\FeatureBundle;
+use yii\base\Event;
+
+class CharacterVariabilityValidation extends FeatureBundle
+{
+    private const PATTERN = '/^(?=.*[0-9])(?=.*[A-Z])(?=.*[a-z])(?=.*[^a-zA-Z0-9]).{8,}$/';
```
gustavs-gutmanis: This checks for a minimum of `8` characters, but we also have a `minLength` property now.

If you set `minLength` to `3`, this will still fail. I suggest we cut this down to min 4, since we need 
* 1 letter
* 1 upper
* 1 lower 
* 1 symbol

What do you think?

seandelaney: Good catch. I was using a password of 8 in my regex testing.

seandelaney: @gustavs-gutmanis Updated

gustavs-gutmanis (dismissed):
Looking great

kjmartens (approved):

PR #1286, opened 2024-05-17
Fix frontend assets for Craft Cloud

timkelty:
These need to be loaded as asset bundle URLs to be correct in a Craft Cloud environment.
See: https://craftcms.com/knowledge-base/cloud-plugin-development#asset-bundles


gustavs-gutmanis:
Addressed in #1299

PR #1288, opened 2024-05-17
remove check for valid form before file uploads

urbantrout:
files should be uploaded no matter if the form itself has validation errors.

### DIFF
--- a/packages/plugin/src/Library/Composer/Components/Form.php
+++ b/packages/plugin/src/Library/Composer/Components/Form.php
@@ -1323,19 +1323,17 @@ abstract class Form implements FormTypeInterface, \JsonSerializable, \Iterator, 
             $isFormValid = false;
         }
 
-        if ($isFormValid) {
-            foreach ($currentPageFields as $field) {
-                if ($field instanceof FileUploadInterface) {
-                    try {
-                        $field->uploadFile();
-                    } catch (\Exception $e) {
-                        $isFormValid = false;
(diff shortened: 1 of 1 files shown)


gustavs-gutmanis:
Imagine a scenario where a persistent but stubborn person keeps clicking "submit" on a form which has an empty required field, but each time they upload a field. 

Or if the form is set to display error messages on spam, which wouldn't send the submission to spam, and a bot kept on uploading a file each time.

These scenarios would fill the server up really fast.

Could you please tell me what the issue is that you're facing in particular?

urbantrout:
Thanks for the quick reply! I see your point.

The thing is: I have a situation where a form has up to 48 file inputs and 2 or 3 text inputs. If the text input does not validate or the user did not upload all files the form gets a validation error and all input fields are cleared. This is very frustrating.
Maybe this behavior is something that should be configurable on a per-form basis.

gustavs-gutmanis:
I feel your pain. Could you switch the form to use ajax instead, that should solve your problem.
Also you could try the **File Drag & Drop** field, which works a bit more robustly.

urbantrout:
Unfortunately that is not possible.

gustavs-gutmanis:
The only thing I would recommend then is to fork the repository and add a patch for this logic. I'm sorry, but there doesn't seem any other way.

urbantrout:
Yeah, I already did that. Thanks anyway!

Issue #1278, opened 2024-05-15
Don't load static assets via PHP (for performance and Cloud compatibility)


timkelty:
Freeform [loads front-end resources via PHP](https://github.com/solspace/craft-freeform/blob/v5/packages/plugin/src/controllers/ResourcesController.php) (`Response::sendFile`), which is very inefficient. This is exacerbated on infrastructure Craft Cloud, where PHP can never send files, and will result in a redirect to and s3 upload.

Static front-end resources should always be defined via Asset Bundles, allowing them be be served directly.

There are some guidelines [here](https://craftcms.com/knowledge-base/cloud-plugin-development#other-best-practices) regarding this, but we will continue to improve them.

gustavs-gutmanis:
Hi @timkelty 

Can I ask you a couple of questions just so I understand completely?

### Current Implementation
Currently, Freeform allows you to pick one of three options

1. serve Freeform JS and CSS files from a static URLs 
    * `your.site/freeform/plugin.js`
    * `your.site/freeform/plugin.css`
  
      This approach uses `$response->sendFile()` as you mentioned, which sends `Cache-Control`, `ETag` and `Last-Modified` headers, so that the browser never reloads these scripts unless they have changed. The files don't actually exist, but the end-user does not see the difference.

2. serve Freeform JS and CSS files using `\Craft::$app->assetManager->getPublishedUrl()`. This approach does what your PR #1286 is trying to do.

3. serve Freeform JS and CSS files inline in the form via `<script>` tags.

### Question
If we're assuming that the best approach is the **2nd** one, would it be sufficient for people to use that via setting it in their settings? (We could default to it from now on)

What harm does the **1st** approach do, given that it is cached by the browser? The file is read by PHP which should be possible even in Cloud, and then it is being sent as a response with the correct headers?

Assuming that this isn't possible with Craft Cloud, people should still be able to use the **2nd** approach and serve the files via published assets, right?

If that's the case, then we could see if Craft Cloud is being used and remove the option from the settings entirely. Would that work?

timkelty:
Certainly!

> If we're assuming that the best approach is the 2nd one, would it be sufficient for people to use that via setting it in their settings?

Yep! I didn't realize there were options for how these were rendered.

> What harm does the 1st approach do, given that it is cached by the browser? The file is read by PHP which should be possible even in Cloud, and then it is being sent as a response with the correct headers?

[...]

PR #1299, opened 2024-05-23
[SFT-1137][#1278]: Asset loading improvements

gustavs-gutmanis:
### Related Ticket Number
[SFT-1137](https://solspace.atlassian.net/browse/SFT-1137)
#1278

### Description
Loading assets with PHP and returning as files via `Response` was an ineffective way to handle asset transport to the front-end.  
And the existing approach with loading assets as files was sub-optimal.

This PR deprecates and removes the `Insert as Pointers` as an option from the settings, and defaults to `Insert as Files` by default for everyone that used either `Insert as Files` or `Insert as Pointers`

- refactor loading of published urls for assets with proper `path` and `filePath` paramters
- fix attribute attachment on inserted script tags
- deprecate `Insert as Pointers`
- default to `Insert as Files`
- remove all registered routes for loading assets as pointers

@timkelty, would it be possible for you to run this branch on a Craft Cloud install and see that everything is set up correctly?

I have taken inspiration from the #1286 PR, and adjusted the code to use your approach.

[SFT-1137]: https://solspace.atlassian.net/browse/SFT-1137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


timkelty:
@gustavs-gutmanis that works great!
Output urls now look something like this:
- https://cdn.craft.cloud/7d4a83b3-5fb0-4394-8f50-820514c95920/builds/9c1cda2f-cbef-4e1b-9271-ac204147dd74/cpresources/Solspace-Freeform-Resources/js/scripts/front-end/plugin/freeform.js?v=1716453346
- https://cdn.craft.cloud/7d4a83b3-5fb0-4394-8f50-820514c95920/builds/9c1cda2f-cbef-4e1b-9271-ac204147dd74/cpresources/Solspace-Freeform-Resources/js/scripts/front-end/captchas/recaptcha/v2-checkbox.js?v=1716453346

FYI - It looks like the description on the settings still needs to updated (still says stuff about the removed option)

timkelty:
@gustavs-gutmanis FYI - The Solspace org on https://console.craftcms.com/ has a sandbox cloud project, so you can deploy and test these yourself as well.
Since we're using Freeform, I'm happy to test as well, just making sure you knew :)

kjmartens (changes requested):
While this appears to work in my testing, I am concerned about anyone that might be depending on the Static URLs (aka `pointers`) approach. I also assume any cached templates may not work right after upgrade?

If we are actually to deprecate the `pointers` approach, it should at least continue to work for some period of time. If you are 100% certain that removing it will be okay for all users, there is one other scenario where this definitely becomes a breaking change:

If setting **Freeform Script Insertion Location** to _None_ and use `{{ freeform.loadFreeformPlugin() }}` manually in your template, this renders as:

``` html
<script type="text/javascript" src="https://site.test/freeform/plugin.js?v=f4ce7639d2f69410e807b6592a2935f7195fd95b" ></script>
<link rel="stylesheet" href="https://site.test/freeform/plugin.css?v=aff9fd861f04fe776cde9d8617f7a414d5d57b72"  />
```

And those URLs no longer exist, therefore breaking this approach.

Is it a lot of work to partially revert to continue to support `pointers` and/or `loadFreeformPlugin`?

gustavs-gutmanis:
> @gustavs-gutmanis FYI - The Solspace org on https://console.craftcms.com/ has a sandbox cloud project, so you can deploy and test these yourself as well. Since we're using Freeform, I'm happy to test as well, just making sure you knew :)

Thanks, I will use that sandbox for testing from now on :)
Your help is greatly appreciated!

kjmartens (approved):

PR #1324, opened 2024-05-31
fix(SFT-1135): Suppress Craft's minimum password length if we have our own minimum password length setting set

seandelaney:


packages/plugin/src/Integrations/Elements/User/EventListeners/PasswordTransform.php
```diff
@@ -2,20 +2,52 @@
...
+
+                // To add our own `newPassword` rules based on the password field type min length
+                /*
+                $event->rules[] = [
+                    ['newPassword'],
+                    'string',
+                    'min' => $this->minLength,
+                    'tooShort' => \Craft::t('yii', 'Password is too short'),
+                ];
+                */
```
gustavs-gutmanis: I guess this is no longer needed?
Should be removed then

gustavs-gutmanis (changes requested):

gustavs-gutmanis (approved):

kjmartens (approved):
Very nice

PR #1326, opened 2024-06-03
feat(SFT-1120): Added ability to set the default 'From Email' and 'From Name' values for new email notification templates

seandelaney:
### Related Jira Ticket

https://solspace.atlassian.net/browse/SFT-1120


packages/plugin/src/templates/_templates/email/default.twig
```diff
@@ -1,6 +1,16 @@
+{% set fromEmail = "{{ craft.app.projectConfig.get('email.fromEmail') }}" %}
+{% if defaultFromEmail is defined and defaultFromEmail|length %}
+    {% set fromEmail = defaultFromEmail %}
+{% endif %}
+
+{% set fromName = "{{ craft.app.projectConfig.get('email.fromName') }}" %}
+{% if defaultFromName is defined and defaultFromName|length %}
+    {% set fromName = defaultFromName %}
+{% endif %}
```
gustavs-gutmanis: Since this would be written into every template file, I would like to have this logic happen outside of Twig, and be instead injected into the template as a string. Otherwise changing the default will change all previous templates as well.

My suggestion would be to read this value when reading the template content, and replacing these values there instead.

`Models\Settings::getEmailTemplateContent()`

```php
    public function getEmailTemplateContent(): string
    {
        $path = __DIR__.'/../templates/_templates/email/default.twig';
        if (!file_exists($path)) {
            throw new FreeformException(
                Freeform::t(
                    'Could not get email template content. Please contact Solspace.'
                )
            );
        }
        
        $email = $this->defaultFromEmail ?: "{{ craft.app.projectConfig.get('email.fromEmail') }}";
        $name = $this->defaultFromName ?: "{{ craft.app.projectConfig.get('email.fromName') }}";

        $content = file_get_contents($path);

        // This example won't work, needs a proper patter, for replacing just the __placeholder__ part
        $content = preg_replace(
           ['{# fromEmail: __placeholder__ #}', '{# fromName: __placeholder__ #}'], 
           [$email, $name], 
           $content
        );

        return $content;
    }
```

packages/plugin/src/templates/_templates/email/default.twig
```diff
@@ -1,6 +1,16 @@
...
+
 {# subject: New submission on your {{ form.name }} form #}
-{# fromEmail: {{ craft.app.projectConfig.get('email.fromEmail') }} #}
-{# fromName: {{ craft.app.projectConfig.get('email.fromName') }} #}
+{# fromEmail: {{ fromEmail }} #}
+{# fromName: {{ fromName }} #}
```
gustavs-gutmanis:
```suggestion
{# fromEmail: __placeholder__ #}
{# fromName: __placeholder__ #}
```

gustavs-gutmanis (commented):
Would this change work for database templates as well?

gustavs-gutmanis (approved):

kjmartens (approved):

Issue #1327, opened 2024-06-04
[5.3.2] Submissions and Spam listings show duplicate forms


croxton:
### What happened?

In the Submission and Spam views the same form may be listed more than once in the sidebar. This issue appeared in the latest update to 5.3.2.
<img width="422" alt="Screenshot 2024-06-04 at 12 51 17" src="https://github.com/solspace/craft-freeform/assets/279127/2838b8e1-3f20-44b9-a3b7-e49718726469">

To check that this wasn't a corruption of my specific freeform installation, I uninstalled / reinstalled the plugin and tried again -  the problem remained. Looking at the database there are only the two forms I created so they are not actually duplicated - this appears to be a display issue.

### How can we reproduce this?

1. Create two or more forms in a fresh install of Freeform 5.3.2
2. View the Submissions or Spam panes.

### Freeform Edition

Pro

### Freeform Version

5.3.2

### Craft Version

5.1.7

### When did this issue start?

After upgrading from older Freeform version

### Previous Freeform Version

5.2.2

croxton:
Update: I rolled back to previous versions and established that the bug was introduced in `5.3.1`.

kjmartens:
Sorry about this @croxton,

We'll have a fix for this out shortly. 🙂

kjmartens:
This should be resolved in Freeform 5.3.3 now. 🙂

PR #1328, opened 2024-06-04
fix(#1327): duplicate form listing in submissions/spam

gustavs-gutmanis:
* fixing duplicate display of the last form in submissions/spam listings


a-solaris (approved):

kjmartens (approved):
I was never able to duplicate this issue, but trusting that this fixes it for you.

PR #1336, opened 2024-06-06
feat(SFT-1159): Added modules to diagnostic page

seandelaney:
### Related Jira Ticket

https://solspace.atlassian.net/browse/SFT-1159


kjmartens (changes requested):
I guess we need a check to see if there are any modules at all...

screenshot_2024-06-06_at_11 43 42_am_720

seandelaney:
@kjmartens 

What modules have you installed? Can you PM the list in Slack?

I have disabled/uninstalled all modules in had in local dev env and this PR still returns "No modules are installed". 

Plugins are skipped over due to `$module instanceof PluginInterface` condition.

kjmartens (approved):

PR #1340, opened 2024-06-11
[SFT-1160]: adding translation support for builder

gustavs-gutmanis:


gustavs-gutmanis:
Now that I think about it, maybe it's worth just adding all of freeform's translations to the js translations in form view? @kjmartens

kjmartens (approved):

PR #1342, opened 2024-06-12
feat(SFT-1128): survey text answers limited to 20 per text field

kjmartens:


kjmartens:
@gustavs-gutmanis This is an improvement visually and very minor for performance (load time became 12s instead of 13s in my test). I'll merge this for now, but I think it's back to the drawing board for performance improvements. 😕

PR #1367, opened 2024-06-24
feat(SFT-1129): Ability to Archive forms

seandelaney:


packages/client/src/app/pages/forms/edit/store/slices/form/index.ts
```diff
@@ -21,6 +21,8 @@ const initialState: FormState = {
   isNew: true,
   settings: {},
   errors: {},
+  archived: false,
```
gustavs-gutmanis: I think if we have `dateArchived` anything other than `null`, we can assume it's archived, and we don't need a boolean for it.

packages/client/src/app/pages/forms/list/archived-item/archived-item.styles.ts
```diff
@@ -0,0 +1,36 @@
+import { colors } from '@ff-client/styles/variables';
+import styled from 'styled-components';
+
+export const Item = styled.li`
+  line-height: 1.4;
+  list-style-type: disc;
+  vertical-align: center;
```
gustavs-gutmanis: I think `vertical-align` gives absolutely nothing outside of a table cell, you would have to use `display: flex` and `align-items: center` instead.

gustavs-gutmanis: But it doesn't seem like it's needed at all here, it looks great out of the box

packages/client/src/app/pages/forms/list/archived/archived.tsx
```diff
@@ -0,0 +1,43 @@
...
+  if (isEmpty) {
+    return;
+  }
+
+  const label =
+    archived &&
+    (isVisible ? 'Hide ' : 'Show ') +
+      archived.length +
+      ' archived form' +
+      (archived.length === 1 ? '' : 's');
```
gustavs-gutmanis: We need to have these translatable. Lets go with plurals at all times in places where there may be more than 1 of items, Craft's translate isn't powerful enough for plurals I think.

```suggestion
  const label = translate('%s archived forms', isVisible ? 'Hide' : 'Show');
```

packages/client/src/app/pages/forms/list/archived/archived.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const Archived: React.FC<Props> = ({ data }) => {
+  const [isVisible, setIsVisible] = useState(false);
+  const toggleVisibility = (): void => setIsVisible(!isVisible);
+
+  const archived = data && data.filter(({ archived }) => archived);
+  const isEmpty = (data && !archived.length) ?? true;
+
+  if (isEmpty) {
+    return;
```
gustavs-gutmanis: Please use `return null;` I have had issues with using `undefined` before and I think react officially asks to return `null` but I can't find it, so that might be bollocks.

```suggestion
    return null;
```

packages/client/src/app/pages/forms/list/list.empty.tsx
```diff
@@ -32,6 +32,8 @@ const generateFormData = (
       color,
     },
   },
+  archived: false,
```
gustavs-gutmanis: Same as above

packages/client/src/app/pages/forms/list/list.sites.tsx
```diff
@@ -48,7 +48,6 @@ export const ListSites: React.FC = () => {
                     current.handle === site.handle && 'sel'
                   )}
                   onClick={(event) => {
-                    console.log('site.handle', site.handle);
```
gustavs-gutmanis, line 51: 👌🏻

packages/client/src/types/forms.ts
```diff
@@ -33,6 +33,8 @@ export type Form = {
     [namespace: string]: SettingsNamespace;
   };
   ownership?: FormOwnership;
+  archived: boolean;
```
gustavs-gutmanis: Pointing to another place

packages/plugin/src/Bundles/Persistence/FormPersistence.php
```diff
@@ -76,6 +76,13 @@ private function update(PersistFormEvent $event, FormRecord $record): void
         $record->name = $payload->settings?->general?->name ?? null;
         $record->handle = $payload?->settings?->general?->handle ?? null;
 
+        $record->archived = false;
+        $record->dateArchived = null;
```
gustavs-gutmanis: I would suggest creating another column in the `freeform_forms` database called `dateArchived` which is `null` by default, and if it has a date in there, then it's archived, if null, then not :)

packages/plugin/src/Form/Form.php
```diff
@@ -492,6 +498,16 @@ public function getUpdatedBy(): ?User
         return User::findOne($this->updatedByUserId);
     }
 
+    public function isArchived(): bool
+    {
+        return $this->archived;
```
gustavs-gutmanis: this can easily return `$this->dateArchived !== null` with that change

packages/plugin/src/Form/Settings/Implementations/GeneralSettings.php
```diff
@@ -132,6 +132,14 @@ class GeneralSettings extends SettingsNamespace
     )]
     public string $color = '';
 
+    #[Section(self::SECTION_GENERAL)]
+    #[Input\Boolean(
+        label: 'Archive this Form',
+        instructions: 'This form will continue to work and be accessible, but it will no longer appear on the Forms dashboard page, and its submissions will not appear in the Submissions index list.',
+        order: 8,
+    )]
+    public bool $archived = false;
```
gustavs-gutmanis: As discussed in the meeting, it might be more appropriate to have this as an action button in the forms list on cards, rather than a toggle in the settings. (And then `unarchive` as a link in the archived list)

packages/plugin/src/controllers/api/FormsController.php
```diff
@@ -31,6 +31,54 @@ public function __construct(
         parent::__construct($id, $module, $config);
     }
 
+    public function actionArchive(int $id): Response
```
gustavs-gutmanis, line 34: I think it would be fine to have this as a single endpoint `POST /api/forms/ID/archive` which would either archive or restore, based on the current state of the form.

Then you would have the same codebase, the only thing changing would be the update on the `dateArchived` which would be either `null` or a date.

packages/client/src/app/pages/forms/list/list.mutations.ts
```diff
@@ -6,6 +6,44 @@ import { useQueryClient } from '@tanstack/react-query';
...
+
+  return useMutation((id) => axios.post(`/api/forms/${id}/restore`), {
+    onMutate: (id) => id,
+    onSuccess: () => {
+      queryClient.invalidateQueries(
+        QKForms.all(getCurrentHandleWithFallback())
+      );
+    },
+  });
+};
```
gustavs-gutmanis: I think you can use the same endpoint for archiving and restoring, thus making it into a toggle.

gustavs-gutmanis (commented):
Great work, last change and we're all set 👌🏻

kjmartens (approved):

PR #1410, opened 2024-07-08
Update freeform.php

FrDH:
"Activa" is a correct translation of "Assets", but it is stuff you own that can be converted into money :) "Bestand" is actually a bit too general (it means "File") but in this context it definitely makes more sense.


kjmartens:
Thank you for this @FrDH!

kjmartens (approved):

PR #1411, opened 2024-07-08
Update freeform.php

FrDH:


kjmartens:
Thank you for this as well @FrDH 🙂

kjmartens (approved):

PR #1465, opened 2024-08-08
Don't set noCache headers on every request

### LINKED ISSUE #1460: Freeform sends no-cache headers for all web requests
### What happened?

`\Solspace\Freeform\Bundles\Form\HiddenInputs\CsrfInput` is instantiated on every request, and in the constructor, it calls `\Craft::$app->getResponse()->setNoCacheHeaders();`: https://github.com/solspace/craft-freeform/blob/v5/packages/plugin/src/Bundles/Form/HiddenInputs/CsrfInput.php#L20-L23

This means if Freeform is enabled, on every request, regardless of the presence of a form or a CSRF input, no-cache headers are sent.

This is bad, but especially problematic for static cache mechanisms that respond to Cache-Control (eg Craft Cloud).

As of Craft 5.2.6/4.10.5, no-cache headers are sent any time a CSRF token is generated/retrieved, so the easiest thing to do would be to just remove this code entirely.

_No response_

1. Install Freeform
2. Look at the response headers for any request

### Freeform Edition

Pro

### Freeform Version

5.5.4

### Craft Version

5.3.1

Fresh install of Freeform

### Previous Freeform Version

_No response_

timkelty:
Fixes https://github.com/solspace/craft-freeform/issues/1460

It also uses Craft's HTML helper to render the input, which allows it to work with the `asyncCsrfInputs` setting: https://craftcms.com/docs/5.x/reference/config/general.html#asynccsrfinputs


packages/plugin/src/Bundles/Form/HiddenInputs/CsrfInput.php
```diff
@@ -17,21 +18,13 @@ public function __construct()
             return;
         }
 
-        if (!\Craft::$app->request->isConsoleRequest) {
-            // Prevent response from being cached with token
-            \Craft::$app->getResponse()->setNoCacheHeaders();
-        }
-
```
timkelty, line 24: Nocache headers are automatically sent when retrieving/generating a CSRF token in Craft 5.2.6/4.10.5

packages/plugin/src/Bundles/Form/HiddenInputs/CsrfInput.php
```diff
@@ -17,21 +18,13 @@ public function __construct()
...
         Event::on(Form::class, Form::EVENT_OUTPUT_AS_JSON, [$this, 'attachToJson']);
     }
 
     public function attachInput(RenderTagEvent $event)
     {
-        $name = \Craft::$app->getConfig()->getGeneral()->csrfTokenName;
-        $token = \Craft::$app->request->csrfToken;
-
-        $event->addChunk('<input type="hidden" name="'.$name.'" value="'.$token.'" />');
+        $event->addChunk(Html::csrfInput());
```
timkelty, line 28: Allows it to work with `asyncCsrfInputs` https://craftcms.com/docs/5.x/reference/config/general.html#asynccsrfinputs

gustavs-gutmanis (dismissed):
Thank you for this!

gustavs-gutmanis:
Would this work with previous craft versions though?

timkelty:
> Would this work with previous craft versions though?

It wouldn't, but I just updated the PR so it will. The important thing is that you move where you're doing it, so it isn't called on every request.

packages/plugin/src/Bundles/Form/HiddenInputs/CsrfInput.php
```diff
@@ -17,25 +18,19 @@ public function __construct()
             return;
         }
 
-        if (!\Craft::$app->request->isConsoleRequest) {
-            // Prevent response from being cached with token
-            \Craft::$app->getResponse()->setNoCacheHeaders();
-        }
```
timkelty, line 23: Calling this here results in no-cache headers sent for _every_ craft request.

gustavs-gutmanis (approved):
Thank you!

PR #1472, opened 2024-08-13
[SFT-1356]: Url Site Parameter - Undefined

gustavs-gutmanis:
* fix `?site=` query parameter when navigating away from the react app


kjmartens (approved):
This seems to be working correctly now, at least when the **Site Filtering for Forms and Submissions** setting is enabled. It doesn't seem to work 100% (keeps defaulting to default/primary site) when this setting is disabled. However, I assume this is a limitation of using Freeform without the Sites feature, so I'll pass it 😅

PR #1475, opened 2024-08-13
feat(SFT-1339): Updated status column styling to match Craft 5 styling for CP submissions/spam indexes

seandelaney:


kjmartens (changes requested):
Noted various issues in Jira:
https://solspace.atlassian.net/browse/SFT-1339?focusedCommentId=22494

kjmartens (approved):

PR #1514, opened 2024-09-09
fix(SFT-1403): added missing OAuth2 interface that stores the refresh token and expiration timestamps

seandelaney:


seandelaney:
Closing as misread API docs. I thought the refresh token was coming back in milliseconds but it's using seconds.

PR #1606, opened 2024-11-05
fix(SFT-1523): possible infinite loop

seandelaney:


packages/plugin/src/Elements/Submission.php
```diff
@@ -86,6 +86,10 @@ public function __construct($config = [])
 
     public function __get($name): mixed
     {
+        if (null === $name) {
```
gustavs-gutmanis: I think it might not be `null`
```suggestion
        if (empty($name)) {
```

gustavs-gutmanis (dismissed):

kjmartens (approved):

PR #1656, opened 2024-11-21
fix(SFT-1594): correctly fetch the correct fields based on the category for Pipedrive integrations

### LINKED ISSUE #1596: Pipedrive integration fails
### What happened?

The Pipedrive integration returns the following error. Currently, only the Map to Lead is enabled, and relevant fields are mapped to Pipedrive.

```shell
Pipedrive: {"success":false,"error":"Scope and URL mismatch","errorCode":403,"error_info":"Please check [developers.pipedrive.com](http://developers.pipedrive.com/)"} {"integration":{"id":4,"handle":"pipedrive"}} []
```

NA

### Freeform Edition

Pro

### Freeform Version

5.6.8

### Craft Version

4.12.5

After upgrading from older Freeform version, After upgrading from older Craft version

### Previous Freeform Version

_No response_

seandelaney:
Fixes https://github.com/solspace/craft-freeform/issues/1596

I'm a little unsure how this previously worked. I know Leads uses `dealFields` endpoint but since we refactored this integration, lead fields were saved under Deal category in the DB. Now Pipedrive or API Integration base class is passing in Deal as category for Leads when fetching fields and therefore not finding any fields, throwing a generic "out of scope" error.

I feel like the API has slightly changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Fixed
 - Fixed a bug where Matrix relation titles would not always display correctly in the **Usage in Elements** tab. Set `Entry ID#` as a title fallback for Matrix relations without a title.
+- Fixed a bug where the Pipedrive integration was not correctly mapping to Lead fields.
 - Fixed a bug where required field validation wasn't working correctly when using `is one of` and `is not one of` conditional rules logic on it.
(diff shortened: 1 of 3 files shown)


kjmartens (approved):
Tested, looks good!

PR #1659, opened 2024-11-26
[SFT-887]: Improvements to Table field type

gustavs-gutmanis:
* adding `radios` and `textarea` field types
* adding better controls for editing option and checkbox types


kjmartens (changes requested):
I'm seeing issues with Radio fields... issues noted in Jira

kjmartens (approved):
👌

PR #1665, opened 2024-11-30
[SFT-1627]: Freeform formatting template resources are not Craft Cloud compatible

gustavs-gutmanis:


seandelaney (commented):
All works fine in non-cloud based projects.

I thought we had a Cloud project already setup but there was none (so I didn't want to start any trials without approval first)

kjmartens (approved):

PR #1718, opened 2025-01-17
[SFT-1698]: Ability to mark specific fields inside Table fields as required

gustavs-gutmanis:


kjmartens (approved):
@gustavs-gutmanis There are a couple of issues, but I will open a new Jira ticket

PR #1752, opened 2025-01-27
[SFT-1739]: Add sqrt function support to calculation fields

thejahid:
Square root support in Calculation field.


packages/plugin/src/Bundles/Fields/Validation/CalculationFieldValidation.php
```diff
@@ -13,6 +13,7 @@
 class CalculationFieldValidation extends FeatureBundle
 {
     private const GET_VARIABLES_PATTERN = '/field:([a-zA-Z0-9_]+)/u';
+    private ?ExpressionLanguage $expressionLanguage = null;
```
gustavs-gutmanis: I don't think you need to have it as optional, since you're instantiating it in the constructor immediately.

```suggestion
    private ExpressionLanguage $expressionLanguage;
```

gustavs-gutmanis (dismissed):
Looks great, with one small comment

gustavs-gutmanis (dismissed):

kjmartens (approved):

PR #1757, opened 2025-01-28
[SFT-1720]: Add detailed logging for Email notifications

gustavs-gutmanis:


kjmartens (approved):

kjmartens (approved):

davorpeic:
Hi @gustavs-gutmanis , can you please check something?
In the latest update (5.9.8) form is not sending due null value passed to getById() method, I'm assuming it can't get notificationId for some reason so it stays as initialised (submission is added, only emails are not sent)
https://github.com/solspace/craft-freeform/blob/v5/packages/plugin/src/Jobs/SendNotificationsJob.php#L98

Revert to 5.9.6 works.

The form was working before, what would be the condition for that value to be set? Maybe Im missing something in the settings?
Thanks

kjmartens:
Hi @davorpeic,

Sorry for the trouble. I am unable to duplicate any issues with this. Are you able to provide us with a screenshot of the Freeform Diagnostics page as well as context about which email notification types you're using and how they are configured? Maybe I am not doing something the same in my testing. 🙂 

Also, preferably create a new GH issue with the above info here:
https://github.com/solspace/craft-freeform/issues

Thanks! 🙂

kjmartens:
Hi again @davorpeic,

We checked into the issue and will release an update shortly. 🙂 

Thanks for bringing it to our attention.

davorpeic:
hey @kjmartens sorry for late feedback, was just gonna open proper issue, sorry for not doing it at first. 
Glad you catched it.

thanks
davor

PR #1796, opened 2025-02-13
[SFT-1747]: Ability to disable the Formatting Template setting in builder

gustavs-gutmanis:


kjmartens (approved):
I am seeing a potential issue with AJAX but have noted it in a new task in Jira

PR #1799, opened 2025-02-14
feat(SFT-1774): added console command to convert form and field handles to camel case or underscores

seandelaney:
Default is camel case:

`php craft freeform/freeform/convert-handles`

But you can convert to underscores too:

`php craft freeform/freeform/convert-handles -c underscores` Using alias
`php craft freeform/freeform/convert-handles --case underscores`

You can also do a dry run to see which form and/or field handles would be converted

`php craft freeform/freeform/convert-handles --dryRun`
`php craft freeform/freeform/convert-handles --case underscores --dryRun`


packages/plugin/src/Commands/FreeformController.php
```diff
@@ -0,0 +1,167 @@
...
+            }
+        }
+    }
+
+    private function convertToCase(string $handle): string
+    {
+        return match ($this->case) {
+            'underscores' => str_replace(' ', '', str_replace('-', '_', $handle)),
+            default => lcfirst(str_replace(' ', '', ucwords(str_replace(['_', '-'], ' ', $handle)))),
+        };
```
gustavs-gutmanis, line 165: We are using the [Stringy](https://github.com/danielstjules/Stringy) library, maybe it would be nice to use it here, since it can convert `camelCaseString` to `camel-case-string`, etc.

gustavs-gutmanis (commented):
Real nice work, I have a recommendation for the conversion logic, but other than that it's fantastic.

kjmartens (approved):

PR #1806, opened 2025-02-19
SFT-1809: Adding ability to hide “Fields from other Forms” area in builder for Limited Users

kjmartens:


gustavs-gutmanis (approved):
Looks correct 👍🏻

PR #1811, opened 2025-02-20
fix(SFT-1807): apply rules with conditions only

seandelaney:
Skips over rules with empty criteria when migrating from Freeform 3 or Freeform 4.

Prevents fields being hidden in templates when invalid/empty rules.

Customers can still delete rules with no conditions within the form builder > rules tab.


packages/plugin/src/migrations/m230101_200000_FF4to5_MigrateData.php
```diff
@@ -76,6 +76,10 @@ private function migrateLayoutData(): void
         $this->renameOldFieldColumns();
 
         foreach ($layouts as $formId => $layoutJson) {
+            $this->pageMap = [];
+            $this->fieldMap = [];
+            $this->submitFieldMap = [];
```
gustavs-gutmanis, line 81: What was the issue here I'm wondering? The uid's should never collide thus resetting this per layout wouldn't matter.

seandelaney: Just general cleanup. We've had a guy with 500+ forms and 7,000+ fields recently.

gustavs-gutmanis: yikes, ok fair enough 😄

gustavs-gutmanis (approved):

kjmartens (approved):

PR #1817, opened 2025-02-25
fix(SFT-1827): encrypted email addresses in form limiting

seandelaney:


packages/plugin/src/Bundles/Form/Limiting/FormLimiting.php
```diff
@@ -129,46 +125,40 @@ private function limitByEmail(FormEventInterface $event): void
...
 
+        // Get all submissions for form
         $elements = Table::ELEMENTS;
         $submissions = Submission::TABLE;
         $submissionsContents = Submission::getContentTableName($form);
 
-        $query = (new Query())
-            ->select(['s.[[id]]'])
+        $submissions = (new Query())
+            ->select(implode(', ', $emailFieldColumnNames))
```
gustavs-gutmanis: Btw, you can pass an array here, it's totally fine.

packages/plugin/src/Bundles/Form/Limiting/FormLimiting.php
```diff
@@ -183,13 +173,29 @@ private function limitByEmail(FormEventInterface $event): void
...
+            ->all()
         ;
 
-        $isPosted = (bool) $query->scalar();
+        $encryptionKey = EncryptionHelper::getKey($form->getUid());
 
-        if ($isPosted) {
+        $submissionEmailValues = [];
+
+        foreach ($submissions as $submission) {
```
gustavs-gutmanis: Maybe it makes sense to use `Submission::find()` to use an element query, which will automatically take care of decryption? Just to have less places where we manually do it.

seandelaney: Good idea

packages/plugin/src/Bundles/Form/Limiting/FormLimiting.php
seandelaney: I will update

packages/plugin/src/Bundles/Form/Limiting/FormLimiting.php
gustavs-gutmanis: Turns out it wasn't 😆

kjmartens (changes requested):
Issues noted in Slack

kjmartens (approved):

PR #1835, opened 2025-03-04
[SFT-1707]: Ability to set template overrides on field option labels

gustavs-gutmanis:


kjmartens (approved):
There may be something in here breaking the Checkboxes limits feature, but I have created a new task for it.

Issue #1855, opened 2025-03-13
Multipage form submit button is disabled


ManiakMill:
### What happened?

Multipage form submit button first step is disabled.

### Errors and Stack Trace (if available)

```shell
freeform.js?v=1741340052:2 Uncaught (in promise) TypeError: n.fields.filter is not a function
    at freeform.js?v=1741340052:2:27079
    at Array.forEach (<anonymous>)
    at re.reload (freeform.js?v=1741340052:2:27043)
    at new re (freeform.js?v=1741340052:2:30673)
    at freeform.js?v=1741340052:2:34092
    at Array.forEach (<anonymous>)
    at e._initHandlers (freeform.js?v=1741340052:2:34043)
    at e.<anonymous> (freeform.js?v=1741340052:2:45507)
    at freeform.js?v=1741340052:2:32594
    at Object.next (freeform.js?v=1741340052:2:32699)
(anonymous) @ freeform.js?v=1741340052:2
(anonymous) @ freeform.js?v=1741340052:2
re @ freeform.js?v=1741340052:2
(anonymous) @ freeform.js?v=1741340052:2
(anonymous) @ freeform.js?v=1741340052:2
(anonymous) @ freeform.js?v=1741340052:2
(anonymous) @ freeform.js?v=1741340052:2
(anonymous) @ freeform.js?v=1741340052:2
(anonymous) @ freeform.js?v=1741340052:2
ae @ freeform.js?v=1741340052:2
(anonymous) @ freeform.js?v=1741340052:2
setInterval
e @ freeform.js?v=1741340052:2
(anonymous) @ freeform.js?v=1741340052:2
(anonymous) @ freeform.js?v=1741340052:2
(anonymous) @ freeform.js?v=1741340052:2
```

### How can we reproduce this?

1. Create new multipage form (more then 3 pages with required fields)
2. Use some of standart templates (ex. Basic Light)
3. Try to submit first step
4. Button is disabled

### Freeform Edition

Pro

### Freeform Version

5.9.16

### Craft Version

Craft Pro 4.14.10

### When did this issue start?

- [x] Unsure
- [ ] Fresh install of Freeform
- [x] After upgrading from older Freeform version
- [ ] After upgrading from older Craft version
- [ ] After a change to hosting configuration

### Previous Freeform Version

5.9.3

ManiakMill:
Any updates?

kjmartens:
Sorry for the trouble @ManiakMill,

[...]

PR #1890, opened 2025-03-26
Update RulesBundle.php

leevigraham:
Reindex array so serialization creates an array not an object

Addresses: Multipage form submit button is disabled #1855


kjmartens (approved):
Thank you @leevigraham! 🙂

PR #1934, opened 2025-04-08
[SFT-1905]: Can we update Form and Field handle conversion to use ASCII transliteration instead of stripping entirely?

gustavs-gutmanis:


kjmartens (changes requested):
Please ensure that it applies to the form handle in the "New Form" dialog as well 🙂

kjmartens (approved):

PR #1954, opened 2025-04-20
[SFT-1933]: Ability to clone sample formatting templates from the control panel

gustavs-gutmanis:


kjmartens (changes requested):
Please prefix custom template clones with `custom-` 🙂

kjmartens (approved):

PR #1988, opened 2025-05-13
fix(SFT-1764): radial chart types as dashboard widgets

### LINKED ISSUE #1981: Cannot see any charts on the dashboard widgets
### What happened?

I have added some dashboard widgets on Freefrom and I am not seeing any charts or graphs for any of the widgets created.

I don't see any errors or console issues and it looks like all the js files for the charts are loading ok. 

I have tested in multiple browsers (Safari and Chrome) and I also see this isse on my local DDV as well as the production server.

Image

1. Create a new chart widget
2. Set the parameters
3. No charts or graphs show

### Freeform Edition

Pro

### Freeform Version

5.10.12

### Craft Version

Craft Team 5.7.4

- [ ] Unsure
- [x] Fresh install of Freeform

### Previous Freeform Version

_No response_

seandelaney:
Fixes: https://github.com/solspace/craft-freeform/issues/1981

- Known bug in Chart.js
- Do not use chart height setting anymore. Adds to the known bug, makes it worse.
- Now Chart.js is only loaded once.
- Charts only load once JS is ready and renders on resize events.
- If canvas size becomes too small, do not attempt to render.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Solspace Freeform Changelog
 
+## 5.10.14 - 2025-05-13
+
+### Added
+- Added a setting to customize the configuration of the TinyMCE **Rich Text** field toolbar.
+
+### Fixed
+- Fixed a bug where `composer.lock` was included in the Freeform package, potentially causing false security warnings from security scanners.
+
+### Security
+- Fixed a vulnerability allowing isolated Twig template rendering in default field values.
+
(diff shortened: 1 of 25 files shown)


kjmartens (changes requested):
Everything is squished flat now... there's like 0 height for the charts.

kjmartens (approved):
Maximus Splendidious!

PR #1991, opened 2025-05-14
fix(SFT-2077): Replaced Crafts dynamic CSRF input with a custom CSRF input

### LINKED ISSUE #1980: Stripe field won't load on multi-page forms when `asyncCsrfInputs=true`
### What happened?

So we have some forms that are multi-page and include a payment field for Stripe on the last page (re: [Place Stripe field on Last Page](https://docs.solspace.com/craft/freeform/v5/integrations/stripe-payments/#place-stripe-field-on-last-page). 

While we are caching a much of the site, we actually don't cache modules that render a form. That said, if we don't set `asyncCsrfInputs` to `true` for the site than Craft will send a no cache header on every request that uses a CSRF token.

We've accommodated this by implementing the form caching instructions noted [here](https://docs.solspace.com/craft/freeform/v5/templates/caching/#static-caching--blitz) with some alterations, but accomplishing the same effect. Difference are as follows:
1. Craft includes a built in action for fetching CSRF info already—no need to create an uncached twig template to render that JSON.  
```js
window.$app.csrf = fetch('/actions/users/session-info', {
    headers: { 'Accept': 'application/json' },
}).then(res => res.json()).then(res => ({
    name: res.csrfTokenName,
    value: res.csrfTokenValue,
}))
```
2. CSRF data is appended to FormData before submit:  
```js
document.addEventListener('DOMContentLoaded', async () => {
    const csrf = await window.$app.csrf
    const forms = document.querySelectorAll('form')

    for (const form of forms) {
        form.addEventListener('freeform-ajax-before-submit', event => {
            event.data.append(csrf.name, csrf.value)
        })
    }
})
```  
That said, we have also tested what is in the docs with the following replacing the `freeform-ajax-before-submit` event listener in the `DOMContentLoaded` handler above:
```js
const csrfInput = form.querySelector(`input[name=${csrf.name}]`)

if (csrfInput instanceof HTMLInputElement) {
    csrfInput.value = csrf.value
}
```
However this change does not fix the issue either.

Multi-page forms in general do appear to work with the above setup, but when a Stripe payment field is included on one of these subsequent pages the call to `/freeform/payments/stripe/payment-intents` returns a `400 Bad Request` response. Logs indicate that the call is unable to verify the CSRF token.

This does appear to be an issue on the JS side of things with Stripe, but including the PHP stack track below just in case.

```shell
2025-05-06 17:02:39 [web.ERROR] [yii\web\HttpException:400] yii\web\BadRequestHttpException: Unable to verify your data submission. in /var/www/html/vendor/yiisoft/yii2/web/Controller.php:221
Stack trace:
#0 /var/www/html/vendor/craftcms/cms/src/web/Controller.php(176): yii\web\Controller->beforeAction()
#1 /var/www/html/vendor/yiisoft/yii2/base/Controller.php(176): craft\web\Controller->beforeAction()
#2 /var/www/html/vendor/solspace/craft-freeform/packages/plugin/src/controllers/BaseApiController.php(15): yii\base\Controller->runAction()
#3 /var/www/html/vendor/yiisoft/yii2/base/Module.php(552): Solspace\Freeform\controllers\BaseApiController->runAction()
#4 /var/www/html/vendor/craftcms/cms/src/web/Application.php(361): yii\base\Module->runAction()
#5 /var/www/html/vendor/yiisoft/yii2/web/Application.php(103): craft\web\Application->runAction()
#6 /var/www/html/vendor/craftcms/cms/src/web/Application.php(329): yii\web\Application->handleRequest()
#7 /var/www/html/vendor/yiisoft/yii2/base/Application.php(384): craft\web\Application->handleRequest()
#8 /var/www/html/web/index.php(12): yii\base\Application->run()
#9 {main} {"memory":7872312,"exception":"[object] (yii\\web\\BadRequestHttpException(code: 0): Unable to verify your data submission. at /var/www/html/vendor/yiisoft/yii2/web/Controller.php:221)"}
```

1. Configure Craft with `asyncCsrfInputs => true`
2. Create a multi-page form with a Stripe field included on any page but the first
3. Submit the first page to render the page including the Stripe field.

### Freeform Edition

Pro

### Freeform Version

5.10.11

### Craft Version

5.7.4

- [ ] Unsure
- [x] After upgrading from older Freeform version
- [x] After upgrading from older Craft version

### Previous Freeform Version

3.13.38

seandelaney:
Fixes https://github.com/solspace/craft-freeform/issues/1980

After deeper investigation, the `asyncCsrfInput` config setting does not render correctly in our multipage handling and using `Html::csrfInput()`

On page 1, it seems to render, meaning `<craft-csrf-input></craft-csrf-input>` gets converted into `<input type="hidden" name="CSRFTokenName" value="CSRFTokenValue" />` as expected.

On pages 2+, `<craft-csrf-input></craft-csrf-input>` is again injected into the page but it does not get converted into `<input type="hidden" name="CSRFTokenName" value="CSRFTokenValue" />`. The JS used by Craft to find and replace the HTML tag, fails.

A workaround for Freeform is to just roll our own CSRF input, whether `asyncCsrfInput` config setting enabled or not. Rolling our own does not affect how Craft handles `asyncCsrfInput` for other non Freeform forms on the page or other security usages. This change just tells Freeform to always use our own approach. To get CSRF token name and value and inject into the form each time the form is rendered.

I set `asyncCsrfInputs` to be true in my Craft config.

I've tested a multiple page form with Stripe and all worked as expected.

Slightly different JS than the user reporting the issue as I felt the CSRF input should be set upon "freeform-ready" event rather than the "freeform-ajax-before-submit" event fired on submit button clicks. When Stripe payment fields are included in the form, setting the CSRF input value on submit button click will not work as expected. Stripe grabs and uses the CSRF input upon page load when sending payment intent requests, which also happen on page load, not on submit button click.

```
<script>
    document.addEventListener("DOMContentLoaded", function () {
        const forms = document.querySelectorAll('[data-id="{{ form.anchor }}"]');
        forms.forEach(form => form.addEventListener("freeform-ready", function(event) {
            const freeform = event.target.freeform;

            event.options.scrollElement = freeform.form;

            fetch('/actions/users/session-info', { headers: { 'Accept': 'application/json' }})
                .then(response => response.json())
                .then(response => {
                    const csrfInput = freeform.form.querySelector('input[name=' + response.csrfTokenName + ']');
                    if (csrfInput) {
                        csrfInput.value = response.csrfTokenValue;
                    }
                });

            /*
            // https://docs.solspace.com/craft/freeform/v5/templates/caching/#static-caching--blitz
            fetch('/dynamic')
                .then(response => response.json())
                .then(response => {
                    if (response.csrf && response.csrf.name && response.csrf.value) {
                        const csrfInput = freeform.form.querySelector('input[name=' + response.csrf.name + ']');
                        if (csrfInput) {
                            csrfInput.value = response.csrf.value;
                        }
                    }
                });
            */
        }));
    });
    </script>
```

image

image


gustavs-gutmanis:
I spent some time and found out that Craft supports a CSRF token passed as a header, instead of being an input field. (Though Yii supports that out of the box, it turns out Craft uses their own header for this)

I implemented an auto-fetch mechanism for CSRF tokens for ajax requests sent from throughout Freeform, which are then added as `X-Craft-Csrf` headers to the requests.
The token is only fetched once per session (which might or might not be optimal, we have to see. But it works like it worked before, no refresh before submitting)

The current approach works in all scenarios and for cached templates out of the box, so there is no more need to fetch and replace csrf tokens by yourself for cached pages.

seandelaney:
@kjmartens 

Tested G's changes without our manual CSRF JS approach with Multiple page form, Stripe and Blitz setup. All worked perfectly.

kjmartens (approved):

PR #1995, opened 2025-05-19
fix(SFT-2076): mailchimp date and birthday field type mappings

seandelaney:


packages/plugin/src/Integrations/EmailMarketing/Mailchimp/EventListeners/MailchimpBirthdayValueProcessor.php
```diff
@@ -0,0 +1,81 @@
...
+        $value = $event->getValue();
+        if (empty($value)) {
+            return;
+        }
+
+        // Mailchimp requires MM/DD for a valid birthday input
+        $event->setValue($this->normalizeToBirthday($value));
+    }
+
+    public function normalizeToBirthday($value): string
```
gustavs-gutmanis: This should be set to `private` as it will and should never be used by code outside of this class.

packages/plugin/src/Integrations/EmailMarketing/Mailchimp/EventListeners/MailchimpBirthdayValueProcessor.php
```diff
@@ -0,0 +1,81 @@
...
+            return;
+        }
+
+        $value = $event->getValue();
+        if (empty($value)) {
+            return;
+        }
+
+        // Mailchimp requires MM/DD for a valid birthday input
+        $event->setValue($this->normalizeToBirthday($value));
```
gustavs-gutmanis, line 61: If the freeform field is a `DatetimeField`, there's a convenient way to get the Carbon object directly and just return a new format of the value.

```
$freeformField = $event->getFreeformField();
if ($freeformField instanceof DatetimeField) {
  $carbon = $freeformField->getCarbon();
  $event->setValue($carbon->format('m/d'));
}
```

This is just a quick example. I would first set the `m/d` as a private `const` in this class, and send this carbon value to the `::normalizeToBirthday()` where it would check if the value is a datetime instance, which would then return the formatted value directly.

Also I think that if you do `new \DateTime($formattedValue);` it automatically figures out the date from the format, so there's no need to manually write all of the formats (of which there are many I think).

packages/plugin/src/Integrations/EmailMarketing/Mailchimp/BaseMailchimpIntegration.php
```diff
@@ -385,8 +385,10 @@ private function fetchContactFields(Client $client, string $listId, string $cate
 
         foreach ($json->merge_fields as $field) {
             $type = match ($field->type) {
-                'text', 'website', 'url', 'dropdown', 'radio', 'date', 'birthday', 'zip' => FieldObject::TYPE_STRING,
+                'text', 'website', 'url', 'dropdown', 'radio', 'zip' => FieldObject::TYPE_STRING,
                 'number', 'phone' => FieldObject::TYPE_NUMERIC,
+                'birthday', => FieldObject::TYPE_BIRTHDAY,
```
gustavs-gutmanis: Lets have the `birthday` type in the `MailchimpIntegrationInterface` class, since it's only used by Mailchimp.

packages/plugin/src/Library/Integrations/DataObjects/FieldObject.php
```diff
@@ -24,6 +24,7 @@ class FieldObject
     public const TYPE_TIMESTAMP = 'timestamp';
     public const TYPE_MICROTIME = 'microtime';
     public const TYPE_BOOLEAN = 'boolean';
+    public const TYPE_BIRTHDAY = 'birthday';
```
gustavs-gutmanis: Since the `birthday` type is very specific to mailchimp, we should keep that type inside the mailchimp class, rather than the `FieldObject` which contains only generic types used by everything.

gustavs-gutmanis (commented):
Very nice, just a couple of suggestions on how to improve it a bit.

gustavs-gutmanis (dismissed):
Fantastic!

kjmartens (changes requested):
Noted issues in Jira task

kjmartens (approved):

PR #2050, opened 2025-07-01
[SFT-2198]: No Caches header issue

gustavs-gutmanis:


kjmartens (changes requested):
If `CRAFT_ASYNC_CSRF_INPUTS=true` + new CSRF Method setting is `Never`, the form fails every time:

>Bad Request: Unable to verify your data submission.

kjmartens (approved):

PR #2055, opened 2025-07-03
feat(SFT-2213): added reason value to spam reason

seandelaney:


packages/plugin/src/Form/Form.php
```diff
@@ -442,19 +442,19 @@ public function isAjaxResetDisabled(): bool
         return $this->disableAjaxReset;
     }
 
-    public function markAsSpam(string $type, string $message): self
+    public function markAsSpam(string $type, string $message, ?string $value): self
```
gustavs-gutmanis: I would suggest setting it to `?string $value = null` so that you don't have to touch any of the previous function calls.

packages/plugin/src/templates/submissions/edit.twig
```diff
@@ -190,7 +190,7 @@
 			</legend>
 			<ul class="spam-reasons">
 				{% for reason in submission.spamReasons %}
-					<li>{{ reason.message|t('freeform') }}</li>
+					<li>{{ reason.message|t('freeform') }}<br><br><i>{{ reason.value|t('freeform') }}</i></li>
```
gustavs-gutmanis: we should probably check if there is a value at all, and only then do the `<br><br>` markup.

gustavs-gutmanis (dismissed):

kjmartens (approved):

PR #2058, opened 2025-07-04
[SFT-2206]: Update Freeform's toggle/switches to match Craft 5.8's new style

gustavs-gutmanis:


kjmartens (changes requested):
There are display issues for Craft 4 and 5.0-5.7 sites. Issue added to Jira

kjmartens (approved):

PR #2072, opened 2025-07-10
[SFT-1749]: add exports for form groups, favorite fields and limited users

gustavs-gutmanis:


kjmartens (changes requested):
There are a couple of issues:

1. When exporting from one site, I only selected a few of my Favorite Fields and only 1 of 2 Limited User profiles. When I import this on another site, it shows me ALL Favorite Fields and ALL Limited User profiles
2. I get the following error on the Import page pertaining to form groups (a few of them successfully imported, but not all):
> SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`craft5install`.`freeform_forms_groups`, CONSTRAINT `freeform_forms_groups_siteId_fk` FOREIGN KEY (`siteId`) REFERENCES `sites` (`id`) ON DELETE CASCADE)

kjmartens (changes requested):
Something broke in a recent update:

> fix(SFT-1749): limit exports to selected only for favorites, groups and limited users

I can no longer see options to select the new features (form groups, favorite fields, limited users).

@gustavs-gutmanis

kjmartens (approved):

PR #2076, opened 2025-07-13
[SFT-2231]: AI ✨

thejahid:
- Implemented AiSummaryService to handle AI summary field processing, including caching and API calls to OpenAI and Gemini.
- Added new AI settings section in the plugin settings for enabling/disabling AI features, selecting providers, and configuring API keys and model parameters.
- Created a new template for AI settings in the admin panel.
- Introduced a new submission field template for displaying AI-generated summaries.


packages/plugin/src/Services/AiSummaryService.php
```diff
@@ -0,0 +1,258 @@
...
+        // Check if AI is enabled globally and API key is configured
+        if (!$settings->aiEnabled || empty($settings->aiApiKey)) {
+            if (!$settings->aiEnabled) {
+                Freeform::getInstance()->logger->getLogger('ai-summary')->warning(
+                    'AI Summary field processing skipped - AI features are disabled globally',
+                    [
+                        'form' => $form->getHandle(),
+                        'field' => $aiSummaryField->getHandle(),
+                    ]
+                );
```
gustavs-gutmanis: I'm not sure we want to log a warning, if somebody has their AI features disabled. At best we could log a debug statement.

packages/plugin/src/Services/AiSummaryService.php
```diff
@@ -0,0 +1,258 @@
...
+            } else {
+                Freeform::getInstance()->logger->getLogger('ai-summary')->warning(
+                    'AI Summary field processing skipped - API key not configured',
+                    [
+                        'form' => $form->getHandle(),
+                        'field' => $aiSummaryField->getHandle(),
+                    ]
+                );
+            }
```
gustavs-gutmanis: This doese make sense!

packages/plugin/src/Services/AiSummaryService.php
```diff
@@ -0,0 +1,258 @@
...
+        $systemPrompt = $aiSummaryField->getSystemPrompt();
+
+        if (empty($systemPrompt)) {
+            $systemPrompt = 'You are an AI assistant that analyzes form submissions. Provide a concise analysis of the provided content.';
+        }
+
+        return $systemPrompt;
+    }
+
+    private function makeHttpRequest(string $url, array $data, array $headers = []): array
```
gustavs-gutmanis: Since we and Craft are using the `Guzzle` library, this could be reduced a bit and called via `$client->post()` instead :)

packages/plugin/src/controllers/api/SettingsController.php
```diff
@@ -79,6 +79,25 @@ public function actionReliability(): Response
         return $this->returnSuccess();
     }
 
+    public function actionAi(): Response
+    {
+        $this->requirePostRequest();
+        $this->saveSettings(
+            [
+                'aiEnabled' => $this->getRequest()->getBodyParam('aiEnabled', false),
```
gustavs-gutmanis: It would be nice to extract `$this->getRequest()` to a variable, since it's used so many times here.

gustavs-gutmanis (dismissed):
Looks great, just a couple of little touches

packages/plugin/src/Integrations/AI/AiIntegrationInterface.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Solspace\Freeform\Integrations\AI;
+
+use Solspace\Freeform\Library\Integrations\IntegrationInterface;
+
+interface AiIntegrationInterface extends IntegrationInterface
```
gustavs-gutmanis: These are going to always be integrations with API calls, so we should extend `ApiIntegrationsInterface` which will automatically add support for checking status, etc.

packages/plugin/src/Integrations/AI/AiIntegrationInterface.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    public function getApiKey(): string;
+
+    public function getModel(): string;
+
+    public function getMaxTokens(): int;
+
+    public function getTemperature(): float;
+
+    public function processAiRequest(string $systemPrompt, string $userContent, array $options = []): string;
```
gustavs-gutmanis, line 17: this should become a `process(Form $form, Client $client): void` and the integration should read the values from its own instance, I think.

Maybe I'm not in the right here, but I assume the form submits, and the AI integration triggers and reads the AI field properties to get the data needed, and reads any other form properties that are needed for the prompts, and then makes the API call using the authorized `$client` instance.

packages/plugin/src/Integrations/AI/Fields/AiField.php
```diff
@@ -0,0 +1,211 @@
...
+    {
+        return $this->integration;
+    }
+
+    public function getSystemPrompt(): string
+    {
+        return $this->systemPrompt;
+    }
+
+    public function getFieldsToProcess(): array
```
gustavs-gutmanis, line 118: You can probably pass a `Form` instance here, and not only find the field handles to look for, but also get their values.

packages/plugin/src/Services/AiService.php
```diff
@@ -0,0 +1,123 @@
...
+namespace Solspace\Freeform\Services;
+
+use Solspace\Freeform\Fields\FieldInterface;
+use Solspace\Freeform\Form\Form;
+use Solspace\Freeform\Freeform;
+use Solspace\Freeform\Integrations\AI\AiIntegrationInterface;
+use Solspace\Freeform\Integrations\AI\Fields\AiField;
+use yii\base\Component;
+
+class AiService extends Component
```
gustavs-gutmanis, line 12: This could be moved to a feature bundle inside the AI integrations bundle.
You can check out the OAuth bundle or Payments bundle, they have their own services, providers, routes and controllers inside the bundle folder.

You can also leave this here, if it's too much trouble, but completely isolating functionality into a single domain makes maintaining it easier and cleaner.

kjmartens (approved):

PR #2082, opened 2025-07-17
Fixes #2081, Multiselect Field options properly translated

### LINKED ISSUE #2081: Multiselect Field Options not translated
### What happened?

For multiselect fields the option gets translated incorrectly. Instead of the translated version of the options they all get the translated label of the field instead.

The issue can be fixed in solspace/craft-freeform/packages/plugin/src/Fields/Implementations/MultipleSelectField.php on line 120-129:

Instead of 
```php
$output .= Html::tag(
    $optionAttributes->getTag('option'),
    $this->translate('label', $option->getLabel()),
    $optionAttributes->toHtmlTagArray([
        'i' => $index,
        'index' => $index,
        'option' => $option,
        'field' => $this,
    ])
);
```

it should be

```php
$output .= Html::tag(
    $optionAttributes->getTag('option'),
    $this->translateOptionLabel($option),
    $optionAttributes->toHtmlTagArray([
        'i' => $index,
        'index' => $index,
        'option' => $option,
        'field' => $this,
    ])
);
```

The code calls for the generic translate method of the field instead of the translateOptionLabel method.
In the Dropdown Class the code is correct.

```shell

```

1. Create a field with a multiselect option
2. Select either a category / entries or custom values as options
3. Translate them
4. Render the form in a different language than the primary one

### Freeform Edition

Pro

### Freeform Version

5.11.5

### Craft Version

5.8.6

- [x] Unsure

### Previous Freeform Version

_No response_

AramLoosman:
I've fixed the line mentioned in the Issue #2081.


kjmartens (approved):
Thank you!

PR #2096, opened 2025-07-24
fix(SFT-2241): strengthen locking in digest service and prevent duplicate digest jobs

seandelaney:
Improved locking by moving the locking inside `DigestService::triggerDigest()`. Checking the lock in `DigestBundle::triggerDigest()`, before queuing the job was too late as once jobs started running, Craft's queue runs synchronously in the same request, and jobs are run immediately and sequentially. This eliminates race conditions so only one `triggerDigest()` fires before a lock is applied and observed. `DigestService::triggerDigest()` still has a lock that prevents multiple jobs filled up the queue causing noise, so not all is lost in that aspect.

We now create a notification log record before sending the email because we were checking if a record exists first, then send the digest, then insert the `NotificationLogRecord`. By the time the log record is saved, another job could have already reached the send stage. This was identified by inspecting server logs from Kelseys brothers server. Also added DB Transaction to sending the email incase it fails so job can be triggered as failed. 

Added unique flag to send digest job to help Craft avoid queuing duplicates and not retry automatically if failed. Now that the real locking is within `DigestService::triggerDigest()`, now if a job was added 30x and 5 failed, this prevents duplicates occurring. Also added `getUniqueId` so we can clearly see duplicates going forward, if duplicates were to occur.

Added unique constraint to `NotificationLogRecord` table to ensure only one record per type per `digestDate`, where `digestDate` is derived from the frequency window (day/week)

Added automatic purging of old notification logs to prevent bloat

Improved logging information

image

image

image

image

in my debug logs screenshot, I triggered new digest emails. I then cleared cache to trigger another round of processing to show it correctly locking and preventing duplicates.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
   "name": "solspace/craft-freeform",
   "description": "The most flexible and user-friendly form building plugin!",
-  "version": "5.11.7",
+  "version": "5.11.7.2",
   "type": "craft-plugin",
   "authors": [
     {
@@ -53,7 +53,7 @@
     "post-update-cmd": "\"vendor/bin/phpcs\" --config-set installed_paths vendor/phpcompatibility/php-compatibility"
   },
   "extra": {
-    "schemaVersion": "5.6.3",
+    "schemaVersion": "5.6.4",
     "handle": "freeform",
     "class": "Solspace\\Freeform\\Freeform",
     "name": "Freeform",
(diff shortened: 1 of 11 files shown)


kjmartens:
@seandelaney I had issues with:

- Handling duplicate date rows (which appeared to randomly happen approximately once a month)
- Running the migration on Postgres

I had ChatGPT suggest changes, which ultimately got both issues resolved in my testing.

I committed the suggested changes I had to make to get this to handle duplicates as well as work on both MySQL and Postgres.

Can you review and clean it up if necessary? 🙂

kjmartens (approved):

PR #2107, opened 2025-08-03
Migrate from Formie

thejahid:


packages/client/src/app/pages/import-export/index.tsx
```diff
@@ -19,6 +19,10 @@ export const ImportExport: React.FC = () => {
       title = 'Import from Express Forms';
       break;
 
+    case '/import/formie3':
```
gustavs-gutmanis: Is it easy to change this to `/import/formie/v3` maybe? Would make the URL look so much prettier

packages/client/src/app/pages/import-export/import/formie3/formie.tsx
```diff
@@ -0,0 +1,136 @@
...
+          }))
+        }
+      />
+
+      <button
+        className={classes(
+          'field btn',
+          'submit',
+          active && 'disabled',
+          !options.forms.length &&
```
gustavs-gutmanis: There's a helper function to check this [here](https://github.com/solspace/craft-freeform/blob/v5/packages/client/src/app/pages/import-export/export/export.operations.ts#L5).

It helps by having all these manual checks in one place, so that if we ever have anything new there, we don't have to find all the places where  this is checked

packages/plugin/src/Bundles/Backup/Export/FormieV3/Fields/FormieFieldMapper.php
```diff
@@ -0,0 +1,109 @@
...
+            new SummaryProcessor(),
+            new TagsProcessor(),
+            new UsersProcessor(),
+            new VariantsProcessor(),
+        ];
+    }
+
+    public function mapField($formField, string $formUid, int $index): ?Field
+    {
+        $processor = $this->findProcessor($formField);
```
gustavs-gutmanis: Nice abstraction, but we could improve it by loading all instances of `FieldProcessorInterface` classes inside the folder.

Take a look at how we're loading feature bundles, we have a class which is able to load specific interface implementations.

What do you think?

This way you would not need to specify them here manually.

packages/plugin/src/Bundles/Backup/Export/FormieV3/Fields/Processors/AbstractFieldProcessor.php
```diff
@@ -0,0 +1,99 @@
...
+        $field->name = $formField->label ?? 'Field '.($index + 1);
+        $field->handle = $this->getFieldHandle($formField->handle ?? 'field'.$index);
+        $field->type = $this->getFreeformFieldClass();
+        $field->required = $formField->required ?? false;
+        $field->metadata = $this->getFieldMetadata($formField);
+
+        return $field;
+    }
+
+    protected function getFieldHandle($currentHandle): string
```
gustavs-gutmanis: we're missing some type safety here

```suggestion
    protected function getFieldHandle(string $currentHandle): string
```

packages/plugin/src/Bundles/Backup/Export/FormieV3/Fields/Processors/AbstractFieldProcessor.php
```diff
@@ -0,0 +1,99 @@
...
+
+namespace Solspace\Freeform\Bundles\Backup\Export\FormieV3\Fields\Processors;
+
+use Solspace\Freeform\Bundles\Backup\DTO\Field;
+use Solspace\Freeform\Bundles\Backup\Export\FormieV3\Fields\Interfaces\FieldProcessorInterface;
+use Solspace\Freeform\Library\Helpers\HashHelper;
+
+abstract class AbstractFieldProcessor implements FieldProcessorInterface
+{
+    public function process($formField, string $formUid, int $index): ?Field
```
gustavs-gutmanis: Is it possible to specify the type of `$formField` here or can it be anything?
If it's anything, lets still add `mixed` as the type.

packages/plugin/src/Bundles/Backup/Export/FormieV3/Fields/Processors/AbstractFieldProcessor.php
```diff
@@ -0,0 +1,99 @@
...
+        ];
+    }
+
+    /**
+     * Default implementation for mapping field options.
+     * Can be overridden by individual processors for custom logic.
+     *
+     * @param mixed $field
+     */
+    protected function mapFieldOptions($field): array
```
gustavs-gutmanis: It kind of feels wrong to have option mapper in the very base `AbstractFieldProcessor`, since options will only be present for some small superset of field types.

I would recommend adding another intermediary class for option fields, which extends `AbstractFieldProcessor` and implements this method, and only use it for processors which process option based fields.

packages/plugin/src/Bundles/Backup/Export/FormieV3/Fields/Processors/AddressProcessor.php
```diff
@@ -0,0 +1,260 @@
...
+use Solspace\Freeform\Fields\Implementations\TextField;
+use Solspace\Freeform\Fields\Properties\Options\Predefined\Types\Countries\Countries;
+use Solspace\Freeform\Fields\Properties\Options\Predefined\Types\States\States;
+use Solspace\Freeform\Library\Helpers\HashHelper;
+
+class AddressProcessor extends AbstractFieldProcessor
+{
+    public function canProcess($formField): bool
+    {
+        return 'verbb\formie\fields\Address' === $formField::class;
```
gustavs-gutmanis: Now it's starting to make sense why we can't use type definitions for `$formField`.
If Formie isn't installed, we would get an error here if we used `Addess::class` directly, right?

packages/plugin/src/Bundles/Backup/Export/FormieV3/FormProcessor.php
```diff
@@ -0,0 +1,88 @@
...
+    private FormieFieldMapper $fieldMapper;
+    private FormieLayoutBuilder $layoutBuilder;
+    private FormieNotificationProcessor $notificationProcessor;
+
+    public function __construct(private PropertyProvider $propertyProvider)
+    {
+        $this->fieldMapper = new FormieFieldMapper();
+        $this->layoutBuilder = new FormieLayoutBuilder($this->fieldMapper);
+        $this->notificationProcessor = new FormieNotificationProcessor();
+    }
```
gustavs-gutmanis: If we're using `Craft::$container->get()` to get the processor instance via the Dependency Injector, you can just specify all of these as dependencies directly.

```suggestion
    public function __construct(
        private PropertyProvider $propertyProvider,
        private FormieLayoutBuilder $layoutBuilder,
        private NotificationProcessor FormieNotificationProcessor,
    ) {
    }
```

The DI container will take care of instantiating everything and pass these itself

packages/plugin/src/Bundles/Backup/Controllers/ImportController.php
```diff
@@ -25,6 +27,19 @@ public function actionExpressForms(): Response
         return $this->asSerializedJson($data, context: ['preserve_empty_objects' => false]);
     }
 
+    public function actionFormie3(): Response
+    {
+        try {
+            $propertyProvider = \Craft::$container->get(PropertyProvider::class);
+            $exporter = new FormieV3Exporter($propertyProvider);
+            $data = $exporter->collectDataPreview();
```
gustavs-gutmanis: When using the Dependency Injection Container you don't need to get the property provider, the injector will take care of it.

```suggestion
        try {
            $exporter = \Craft::$container->get(FormieV3Exporter::class);
            $data = $exporter->collectDataPreview();
```

gustavs-gutmanis (commented):
Great work, adding a couple comments for you to consider

PR #2126, opened 2025-08-20
[SFT-2131]: Add filtering options for element Status for Element Source population of fields

gustavs-gutmanis:


kjmartens (changes requested):
Found an issue. Noted in Jira:
https://solspace.atlassian.net/browse/SFT-2131?focusedCommentId=30361

kjmartens (approved):

PR #2154, opened 2025-09-03
[SFT-2066]: Conditional logic on Integrations

gustavs-gutmanis:


kjmartens (changes requested):
Conditional Rules should only be present on the following types:

- ✅ CRM
- ✅ Email Marketing
- Elements (❌ currently doesn't work in my testing)
- Webhooks (❌ currently doesn't work in my testing)
- Other (haven't been able to test yet)

I notice these settings appear on the following types, but shouldn't (as it doesn't exactly make sense):

- Payment Gateways
- Captchas
- Spam Blocking
- Singles
- AI

Trivial issue, but the Loading state is a little goofy looking:
Image

kjmartens (approved):

PR #2162, opened 2025-09-08
[SFT-2248]: Added ai spam protection functionality

thejahid:


packages/plugin/src/Integrations/SpamBlocking/AI/AiSpamProtection.php
```diff
@@ -0,0 +1,399 @@
...
+
+        if (str_contains($normalized, 'NOT_SPAM')
+            || str_contains($normalized, 'NOT APPEAR TO BE SPAM')
+            || str_contains($normalized, 'DOES NOT APPEAR TO BE SPAM')
+            || str_contains($normalized, 'IS NOT SPAM')
+            || str_contains($normalized, 'LEGITIMATE')
+            || str_contains($normalized, 'STRAIGHTFORWARD REQUEST')
+        ) {
+            return 'NOT_SPAM';
+        }
```
gustavs-gutmanis: I think it would be better to check a direct comparison `=== 'NOT_SPAM'` etc.
`IS NOT SPAM` contains `NOT SPAM`. I see that you use `_` between the letters for that single thing, but why not use strict comparisons instead?
Then we can use `match` or `switch`

packages/plugin/src/Integrations/SpamBlocking/AI/AiSpamProtection.php
```diff
@@ -0,0 +1,399 @@
...
+
+        if (preg_match('/\{[^}]*"reason"[^}]*\}/', $response, $matches)) {
+            $jsonData = json_decode($matches[0], true);
+            if (\JSON_ERROR_NONE === json_last_error() && isset($jsonData['reason'])) {
+                return trim($jsonData['reason']);
+            }
+        }
+
+        return null;
+    }
```
gustavs-gutmanis: This json check logic repeats a couple of times, maybe we can write the logic in a single private function that gets called instead?

packages/plugin/src/Integrations/AI/Gemini/Versions/GeminiV1Beta.php
```diff
@@ -59,15 +59,18 @@ public function processAiRequest(string $systemPrompt, string $userContent, arra
         ]);
```
gustavs-gutmanis: Here we're using a `new Client()`. The credentials should be appended when using the `clientProvider`. I assume it's not possible to get the client provider inside of the other integration, is that the problem?

packages/plugin/src/controllers/api/IntegrationsController.php
```diff
@@ -257,6 +257,28 @@ public function actionDelete(int $id): Response
         return $this->asEmptyResponse(204);
     }
 
+    protected function get(): array|object
```
gustavs-gutmanis, line 260: Where is this used?

kjmartens:
@thejahid can you resolve conflicts for this PR? And can you rename the integration and files as "AI Spam Analysis"? 🙂

kjmartens (approved):

PR #2177, opened 2025-09-22
[SFT-2289]: PayPal Integration

thejahid:


packages/plugin/src/Integrations/PaymentGateways/PayPal/Controllers/BasePayPalController.php
```diff
@@ -0,0 +1,65 @@
...
+
+        $integration = null;
+        foreach ($integrations as $int) {
+            if ($int->getId() === $integrationId) {
+                $integration = $int;
+
+                break;
+            }
+        }
+
```
gustavs-gutmanis: You can set the `enabled` argument to `true` to search only for enabled integrations by default.

There is a `filter` callable argument for the method, which you can use to narrow the results directly:

```suggestion
        $integration = $this->getIntegrationsService()->getFirstForForm(
            $form,
            Type::TYPE_PAYMENT_GATEWAYS,
            true,
            filter: fn ($integration) => $integration->getId() === $integrationId,
        );
```

packages/plugin/src/Integrations/PaymentGateways/PayPal/Controllers/OrdersController.php
```diff
@@ -0,0 +1,110 @@
...
+        private PayPalOrderService $orderService,
+        private PayPalPriceService $priceService,
+    ) {
+        parent::__construct($id, $module, $config);
+    }
+
+    public function actionCreate(): Response
+    {
+        try {
+            [$form, $integration, $field, $hash] = $this->getRequestItems(handleFormRequest: false);
```
gustavs-gutmanis: `handleFormRequest` is `false` by default already

```suggestion
            [$form, $integration, $field, $hash] = $this->getRequestItems();
```

packages/plugin/src/Integrations/PaymentGateways/PayPal/Controllers/OrdersController.php
```diff
@@ -0,0 +1,110 @@
...
+
+            // Validate integration
+            if (!$integration->isEnabled()) {
+                throw new \Exception('PayPal integration is not enabled');
+            }
```
gustavs-gutmanis: By searching for `enabled` integrations in request items, this check becomes unnecessary

packages/plugin/src/Integrations/PaymentGateways/PayPal/Controllers/OrdersController.php
```diff
@@ -0,0 +1,110 @@
...
+                    'field' => $field,
+                ]
+            )
+        ;
+
+        // Allow dynamic amount: merge posted values (from client) into form before calculating
+        $raw = $this->request->getRawBody();
+        $json = $raw ? json_decode($raw, true) : null;
+        $values = \is_array($json) && isset($json['values']) && \is_array($json['values']) ? $json['values'] : null;
+        if ($values) {
```
gustavs-gutmanis: Using so many conditional checks while inside a ternary operator (`?:`) makes it very hard to read for people: [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity)

```suggestion
        $raw = $this->request->getRawBody();
        $json = json_decode($raw, true);
        $values = $json['values'] ?? null;
        if (\is_array($values)) {
```

packages/plugin/src/Integrations/PaymentGateways/PayPal/Controllers/OrdersController.php
```diff
@@ -0,0 +1,110 @@
...
+            foreach ($form->getLayout()->getFields()->getIterator() as $f) {
+                $handle = $f->getHandle();
+                if (\array_key_exists($handle, $values)) {
+                    $f->setValue($values[$handle]);
+                }
+            }
```
gustavs-gutmanis: Please refrain from using `$f`, `$a`, `$b` and whatever else shorthands anywhere. It may make sense to use for these quick loops, but it is never worth it in the long run.

Also - `::getFields()` returns a `FieldCollection` which is an iterable object and you don't need to access the iterator for iterating over it.

```suggestion
            foreach ($form->getLayout()->getFields() as $formField) {
                $handle = $formField->getHandle();
                if (\array_key_exists($handle, $values)) {
                    $formField->setValue($values[$handle]);
                }
            }
```

packages/plugin/src/Integrations/PaymentGateways/PayPal/Services/PayPalPriceService.php
```diff
@@ -0,0 +1,55 @@
...
+                $amount = $this->getDynamicAmount($form, $field);
+
+                break;
+
+            case PayPalField::AMOUNT_TYPE_FIXED:
+            default:
+                $amount = $field->getAmount();
+
+                break;
+        }
```
gustavs-gutmanis: The initial `$amount = $field->getAmount();` is a useless call, since it gets overwritten by the same thing in the `default` case.

Can be changed to a `match` expression, looking cleaner and shorter, more readable.

```suggestion
        $amount = match ($field->getAmountType()) {
            PayPalField::AMOUNT_TYPE_DYNAMIC => $this->getDynamicAmount($form, $field),
            default => $field->getAmount(),
        };
```

packages/plugin/src/Integrations/PaymentGateways/PayPal/Services/PayPalPriceService.php
```diff
@@ -0,0 +1,55 @@
...
+        $finalAmount = (int) round($amount * 100);
+
+        if ($finalAmount <= 0) {
+            throw new \Exception("PayPal amount must be greater than 0. Current amount: {$amount} (final: {$finalAmount})");
+        }
+
+        return $finalAmount;
+    }
+
+    private function getDynamicAmount(Form $form, PayPalField $field): float
```
gustavs-gutmanis: `$form` is never used.

packages/plugin/src/Integrations/PaymentGateways/PayPal/Services/PayPalPriceService.php
```diff
@@ -0,0 +1,55 @@
...
+                $amount = (float) $value;
+            } elseif (\is_string($value)) {
+                $amount = (float) preg_replace('/[^0-9.]/', '', $value);
+            } elseif (\is_array($value)) {
+                $flat = array_filter(array_map(static fn ($v) => is_numeric($v) ? (float) $v : 0, $value));
+                $amount = (float) array_sum($flat);
+            }
+        }
+
+        return (float) $amount;
```
gustavs-gutmanis: - `$form` is never used.
- `$field` cannot be null, so no need to check if it's present and return 0 just yet.
- the less nested IFs are used, the easier the code is to read and maintain down the road. this can be solved by returning the result in each conditional separately, and avoiding `else` statements. 
- if you're doing nested function calls, split them in logical newline blocks, otherwise it's near impossible to figure out what is happening when reading it.

```suggestion
    private function getDynamicAmount(PayPalField $field): float
    {
        $value = $field->getAmountField()->getValue();
        if (is_numeric($value)) {
            return (float) $value;
        }

        if (\is_string($value)) {
            return (float) preg_replace('/[^0-9.]/', '', $value);
        }

        if (\is_array($value)) {
            return (float) array_sum(
                array_filter(
                    array_map(
                        static fn ($val) => is_numeric($val) ? (float) $val : 0,
                        $value
                    )
                )
            );
        }

        return 0;
    }
```

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+  style?: Record<string, unknown>;
+  createOrder: () => Promise<string> | string;
+  onApprove: (data: { orderID: string }) => Promise<void> | void;
+  onError?: (err: unknown) => void;
+  onCancel?: (_data: unknown) => void;
+};
+
+interface PayPalSDK {
+  Buttons: (config: PayPalButtonsConfig) => PayPalButtons;
+}
```
gustavs-gutmanis, line 10: You should `npm install @paypal/paypal-js` to get the SDK and import that in the script.
These then become obsolete, and you can do `import type { PayPalNamespace as PayPalSDK } from '@paypal/paypal-js';`

This also allows you to use their script loader, where you can pass all of the options for currency, intent, components, etc.

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+  clientId: string;
+  sandbox: boolean;
+  currency?: string;
+  finalizeOnSubmit?: boolean;
+};
+
+const Q = {
+  root: '[data-freeform-paypal-buttons]',
+  orderInput: '[data-freeform-paypal-order]',
+};
```
gustavs-gutmanis: Again, please do not use single letters, instead name the variable so that it makes sense to other people reading it.

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+const DEBUG = (window as Window & { FREEFORM_DEBUG?: boolean }).FREEFORM_DEBUG === true;
+
+function logError(...args: unknown[]) {
+  if (DEBUG) {
+    console.error('PayPal:', ...args);
+  }
+}
+
+function showNotice(root: Element, text: string) {
+  const container = root.closest('[data-field-container="payPalPayment"]') || root.parentElement;
```
gustavs-gutmanis: the `data-field-container` will have the `handle` of the field, so if you name it differently, it will have a different name.
You should be looking for `data-field-type="paypal"` instead.

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+  try {
+    return JSON.parse(attr);
+  } catch {
+    return null;
+  }
+}
+
+type Values = Record<string, string | number | boolean | File>;
+
+type CreateOrderResult = { id: string; status?: string; approve?: string } | null;
```
gustavs-gutmanis: For readability, lets have each property in a new line, and have a newline at the end

```suggestion
type CreateOrderResult = { 
  id: string; 
  status?: string; 
  approve?: string 
};

```

Also lets not have the type include `null`, but have the `null` in the Promise which uses this type.

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+async function createOrder(endpoint: string, integration: string, values?: Values): Promise<CreateOrderResult> {
```
gustavs-gutmanis:
```suggestion
async function createOrder(endpoint: string, integration: string, values: Values): Promise<CreateOrderResult | null> {
```

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+  try {
+    const res = await fetch(endpoint, {
```
gustavs-gutmanis, line 75: I have written a lightweight `ajax` wrapper for doing XHRRequests similar to how `axios` handles it, without using fetch. 

You can check it out in `@lib/plugin/helpers/ajax` and see how it's used in `components/front-end-payments/stripe/elements.queries.ts`

But its not necessary, you can keep fetch if you don't find it easier with the ajax wrapper.

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+
+    const data = (await res.json()) as { id?: string; status?: string; approve?: string };
+    return data && data.id ? { id: String(data.id), status: data.status, approve: data.approve } : null;
+  } catch (error) {
+    logError('Network error during order creation:', error);
+    return null;
+  }
+}
+
+type CaptureOrderResult = { status: string; id: string } | null;
```
gustavs-gutmanis: Same as above.

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+    };
+
+    script.onerror = () => {
+      logError('Failed to load SDK');
+      reject(new Error('Failed to load PayPal SDK'));
+    };
+
+    document.head.appendChild(script);
+  });
+}
```
gustavs-gutmanis: Paypal SDK in `@paypal/paypal-js` has a `loadScript()` function, which accepts all of these parameters. See if you can use that instead.

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+    return;
+  }
+
+  // Initialize PayPal buttons
+  try {
+    // Load SDK first
+    await loadPayPalSDK(config.clientId, config.sandbox, config.currency || 'USD');
+
+    // Wait a bit for the SDK to be fully ready
+    await new Promise((resolve) => setTimeout(resolve, 100));
```
gustavs-gutmanis: We should never rely on timers to wait for something which relies on how fast your internet speed is.
The `await loadScript` should let the script continue only when it is ready.
The paypal sdk loader should handle that, but if not then we need to instead attach a `.then()` to the promise on script loaded, that executes the script which loads buttons.

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+
+    const p = (window as unknown as { paypal?: PayPalSDK }).paypal;
```
gustavs-gutmanis: No single letter variables please.

gustavs-gutmanis: You should have a separate function which handles fetching and returning the single SDK instance, which you call from here via an `await` call, so that it does not continue executing until the SDK is returned.

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+      createOrder: async () => {
+        // Create order
+
+        try {
+          // Serialize current form values so server can compute dynamic amount before order create
+          const fd = new FormData(form);
+          const values: Values = {};
+          fd.forEach((v, k) => {
+            (values as Record<string, unknown>)[k] = v as unknown as string | number | boolean | File;
+          });
```
gustavs-gutmanis: Pretty unreadable with the single letter variables.

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+  const attr = el.getAttribute('data-config');
+  if (!attr) return null;
+  try {
+    return JSON.parse(attr);
+  } catch {
+    return null;
+  }
+}
+
+type Values = Record<string, string | number | boolean | File>;
```
gustavs-gutmanis: Later on you specify the record value type again manually. This can be avoided by defining a type for it here.
Also since it's using `FormData` you can type it as `FormDataEntryValue`.

```suggestion
type Values = Record<string, FormDataEntryValue>;
```

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
gustavs-gutmanis: Checking the code later on, I see that `values` is always defined as at least an empty object of `Record<string, FormDataEntryValue>`, thus you can remove the optional call here, and `JSON.stringify` it always.

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+async function createOrder(endpoint: string, integration: string, values?: Values): Promise<CreateOrderResult> {
+  try {
+    const res = await fetch(endpoint, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        'FF-PAYPAL-INTEGRATION': integration,
+      },
+      credentials: 'same-origin',
+      body: values ? JSON.stringify({ values }) : undefined,
```
gustavs-gutmanis:
```suggestion
      body: JSON.stringify({ values }),
```

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+  } catch (error) {
+    logError('Network error during order capture:', error);
+    return null;
+  }
+}
+
+// Load PayPal SDK dynamically
+function loadPayPalSDK(clientId: string, sandbox: boolean, currency: string = 'USD'): Promise<void> {
+  return new Promise((resolve, reject) => {
+    if ((window as Window & { paypal?: PayPalSDK }).paypal) {
```
gustavs-gutmanis: You're making this check with the same type in 3 places, it should be typed once and reused then.
But a better approach would be to declare the global namespace to have it.
You would add this to the beginning of this file, and then not have to typecast `window` every time.
```ts
declare global {
  interface Window {
    paypal?: PayPalNamespace;
    FREEFORM_DEBUG?: boolean;
  }
}
```

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
gustavs-gutmanis: You have some values typed already and don't need to re-type them here.

```suggestion
          const formData = new FormData(form);
          const values: Values = {};
          formData.forEach((value, key) => {
            values[key] = value;
          });
```

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+        // Payment cancelled
+      },
+    }).render(root);
+  } catch (error) {
+    logError('Failed to initialize buttons:', error);
+  }
+}
+
+document.addEventListener('DOMContentLoaded', async () => {
+  const roots = document.querySelectorAll(Q.root);
```
gustavs-gutmanis: Use `HTMLElement` types when querying html elements, as this makes certain available which are not present in `Element` alone.

```suggestion
  const roots = document.querySelectorAll<HTMLElement>(Q.root);
```

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -0,0 +1,334 @@
...
+  const firstClientId = Array.from(clientIds)[0];
+  const firstConfig = parseConfig(roots[0]);
+  const loadedCurrency = (firstConfig?.currency || 'USD').toUpperCase();
+  if (currencies.size > 1) {
+    logError(
+      'Multiple currencies detected on the same page. Using',
+      loadedCurrency,
+      'for SDK. Other currencies will be skipped.'
+    );
+  }
```
gustavs-gutmanis: Shouldn't we load one SDK per `clientId`? Then we could allow different currencies and different client id's.

packages/plugin/src/Integrations/PaymentGateways/PayPal/EventListeners/FinalizePayment.php
```diff
@@ -0,0 +1,186 @@
...
+            if (!$integration) {
+                continue;
+            }
+
+            $payment = $this->findOrCreatePaymentRecord($submission, $field, $orderId, $integration);
+
+            // Try to fetch capture details to enrich the record
+            try {
+                $client = $this->clientProvider->getAuthorizedClient($integration);
+                $accessToken = $this->fetchAccessToken($client, $integration);
```
gustavs-gutmanis: Access token should probably be fetched during `::getAuthorizedClient`.  
Optimally, we would fetch the token, save it in the database, save the `expiresIn` and the current timestamp. And during the next `::getAuthorizedClient` call re-use the access token we have stored unless it is expired, in which case we would fetch a new one.

Also checking for a `HTTP: 401` error, which would mean we need to re-authorize again.

packages/plugin/src/Integrations/PaymentGateways/PayPal/EventListeners/FinalizePayment.php
```diff
@@ -0,0 +1,186 @@
...
+            ],
+            'form_params' => [
+                'grant_type' => 'client_credentials',
+            ],
+        ]);
+
+        $json = json_decode($response->getBody()->getContents());
+
+        return (string) ($json->access_token ?? '') ?: null;
+    }
```
gustavs-gutmanis: There are two such methods, one in each controller. That alone means it should be taken out of here.
It should happen in the `::getAuthorizedClient`. 
And the client should return the guzzle client which has the access token added to the headers already.

packages/plugin/src/Integrations/PaymentGateways/PayPal/EventListeners/FinalizePayment.php
```diff
@@ -0,0 +1,186 @@
...
+
+    private function fetchOrderWithRetry($client, PayPal $integration, string $orderId, string $accessToken): array
```
gustavs-gutmanis: You shouldn't need to pass the `$accessToken` here.

packages/plugin/src/Integrations/PaymentGateways/PayPal/EventListeners/RenderTemplates.php
```diff
@@ -0,0 +1,134 @@
...
+
+            return;
+        }
+
+        $template = file_get_contents(__DIR__.'/../Templates/submission-field.twig');
+        $metadata = json_decode($payment->metadata ?? '{}', true);
+
+        // Prefer metadata for amount/currency/status; fallback to record
+        $unit = $metadata['unit'] ?? 'dollars';
+        $rawAmount = isset($metadata['amount']) ? (float) $metadata['amount'] : (float) ($payment->amount ?? 0);
```
gustavs-gutmanis: This part is pretty unreadable. In this case an `if/else` makes it more readable.

```suggestion
        $rawAmount = 0;
        if (isset($metadata['amount'])) {
            $rawAmount = (float) $metadata['amount'];
        } else if (isset($payment->amount)) {
            $rawAmount = $payment->amount;
        }
```

packages/plugin/src/Integrations/PaymentGateways/PayPal/EventListeners/RenderTemplates.php
```diff
@@ -0,0 +1,134 @@
...
+        $displayAmount = $rawAmount > 0
+            ? ('cents' === $unit ? number_format($rawAmount / 100, 2) : number_format($rawAmount, 2))
+            : null;
```
gustavs-gutmanis: This should also be split up, there's two ternary operators. That's very hard to read when just reading a PR.

The most important part of code is maintainability and readability. Eventually someone will be working on this codebase, and it should be as easy as possible for them to understand what's going on. That's why splitting code up in small chunks of function calls and objects with clear names is very potent.

 ```suggestion
        $displayAmount = null;
        if ($rawAmount > 0) {
            $isCents = 'cents' === $unit;
            $divider = $isCents ? 100 : 1;
            $displayAmount = number_format($rawAmount / $divider, 2);
        }
```

File: packages/plugin/src/Integrations/PaymentGateways/PayPal/EventListeners/RenderTemplates.php [hunk 30]
```diff
@@ -0,0 +1,134 @@
...
+
+        $displayStatus = isset($metadata['status']) ? (string) $metadata['status'] : (string) $payment->status;
```
gustavs-gutmanis:
```suggestion
        $displayStatus = (string) ($metadata['status'] ?? $payment->status);
```

packages/plugin/src/Integrations/PaymentGateways/PayPal/Fields/PayPalField.php
```diff
@@ -0,0 +1,173 @@
...
+    name: 'PayPal Payment',
+    typeShorthand: 'paypal',
+    iconPath: __DIR__.'/../icon.svg',
+    previewTemplatePath: __DIR__.'/../Templates/paypal-field-preview.ejs',
+)]
+class PayPalField extends AbstractField implements PaymentFieldInterface
+{
+    public const AMOUNT_TYPE_FIXED = 'fixed';
+    public const AMOUNT_TYPE_DYNAMIC = 'dynamic';
+    #[Required]
```
gustavs-gutmanis, line 34: Needs a newline here.

```suggestion
    public const AMOUNT_TYPE_DYNAMIC = 'dynamic';
    
    #[Required]
```

packages/plugin/src/Integrations/PaymentGateways/PayPal/Fields/PayPalField.php
```diff
@@ -0,0 +1,173 @@
...
+        $inputAttributes = (new Attributes())
+            ->set('data-freeform-paypal-order')
+            ->set('name', $this->getHandle())
+            ->set('type', 'hidden')
+            ->set('value', $this->getValue())
+        ;
+        $output .= '<input'.$inputAttributes.' />';
+
+        $provider = \Craft::$container->get(FormIntegrationsProvider::class);
+        $integrations = $provider->getForForm($this->getForm(), PayPal::class);
```
gustavs-gutmanis: These are unused and do a query, they should be removed.

```suggestion
```

packages/plugin/src/Integrations/PaymentGateways/PayPal/PayPal.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public function getClientSecret(): ?string
+    {
+        return $this->getProcessedValue($this->clientSecret);
+    }
+
+    public function isSandbox(): bool
+    {
+        return (bool) $this->useSandbox;
```
gustavs-gutmanis: No need to cast it, since it cannot be anything else

```suggestion
        return $this->useSandbox;
```

packages/plugin/src/Integrations/PaymentGateways/PayPal/PayPal.php
```diff
@@ -0,0 +1,113 @@
...
+            ],
+            'form_params' => [
+                'grant_type' => 'client_credentials',
+            ],
+        ]);
+
+        [, $json] = $this->getJsonResponse($response);
+
+        return (string) ($json->access_token ?? '');
+    }
```
gustavs-gutmanis: This should not exist here, it should happen in the `::getAuthorizedClient` via an event listener.

packages/plugin/src/Integrations/PaymentGateways/PayPal/PayPal.php
```diff
@@ -0,0 +1,113 @@
...
+    {
+        return (bool) $this->useSandbox;
+    }
+
+    public function checkConnection(Client $client): bool
+    {
+        $token = $this->fetchAccessToken($client);
+
+        return !empty($token);
+    }
```
gustavs-gutmanis: Instead, this should query some endpoint (for instance [webhooks](https://developer.paypal.com/docs/api/webhooks/v1/#webhooks_list)) and see if it returns `HTTP: 200`

```suggestion
    public function checkConnection(Client $client): bool
    {
        $response = $client->get($this->getEndpoint('/v1/notifications/webhooks'), [
            'query' => ['page_size' => 1],
        ]);

        return $response->getStatusCode() === 200;
    }
```

packages/plugin/src/Integrations/PaymentGateways/PayPal/Services/PayPalOrderService.php
```diff
@@ -0,0 +1,137 @@
...
+    public function __construct(
+        private IntegrationClientProvider $clientProvider,
+        private IntegrationLoggerProvider $loggerProvider,
+    ) {}
+
+    public function createOrder(Form $form, PayPal $integration, PayPalField $field, string $description, string $hash, int $amount, string $currency): array
+    {
+        try {
+            $client = $this->clientProvider->getAuthorizedClient($integration);
+            $token = $this->fetchAccessToken($client, $integration);
```
gustavs-gutmanis: No token fetching here

packages/plugin/src/Integrations/PaymentGateways/PayPal/Services/PayPalOrderService.php
```diff
@@ -0,0 +1,137 @@
...
+                'application_context' => [
+                    'user_action' => 'PAY_NOW',
+                ],
+            ];
+
+            $response = $client->post($endpoint, [
+                RequestOptions::HEADERS => [
+                    'Authorization' => 'Bearer '.$token,
+                    'Content-Type' => 'application/json',
+                ],
```
gustavs-gutmanis: Headers would be set by the client fetcher.
Very nice approach with the `RequestOptions::` use here, that's nice!

packages/plugin/src/Integrations/PaymentGateways/PayPal/Services/PayPalOrderService.php
```diff
@@ -0,0 +1,137 @@
...
+            ]);
+
+            $json = json_decode($response->getBody()->getContents());
+
+            return (string) ($json->access_token ?? '');
+        } catch (\Exception $e) {
+            throw $e;
+        }
+    }
```
gustavs-gutmanis: Same as before

gustavs-gutmanis (changes requested):
While this seems to work, there are lots of issues to be addressed for the final product. 
Please see all of the attached comments.

kmk142789 (approved):

packages/plugin/src/Integrations/PaymentGateways/PayPal/EventListeners/RegisterField.php
```diff
@@ -36,6 +36,12 @@ public function __construct(
             Form::EVENT_COLLECT_SCRIPTS,
             [$this, 'collectScripts']
         );
+
+        // Register PayPal client provider for client credentials flow
+        new PayPalClientProvider();
+
+        // Register PayPal error handler for 401 errors
+        new PayPalErrorHandler();
```
gustavs-gutmanis: You don't need to do this. Just have those two classes extend the `FeatureBundle` and they will be auto-wired and loaded by Freeform.

packages/plugin/src/Integrations/PaymentGateways/PayPal/PayPal.php
```diff
@@ -48,26 +47,86 @@ class PayPal extends PaymentGatewayIntegration
...
+                'auth' => [$this->getClientId(), $this->getClientSecret()],
+                'form_params' => [
+                    'grant_type' => 'client_credentials',
+                ],
+            ]);
+
+            return 200 === $response->getStatusCode();
+        } catch (\Exception $e) {
+            return false;
+        }
```
gustavs-gutmanis: This should not be in a `try/catch` clause, because if it throws an exception, it is read and the error is displayed to the user.

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -130,44 +129,26 @@ async function captureOrder(
...
-
-    script.onerror = () => {
-      logError('Failed to load SDK');
-      reject(new Error('Failed to load PayPal SDK'));
-    };
-
-    document.head.appendChild(script);
+async function getPayPalSDK(clientId: string, sandbox: boolean, currency: string): Promise<PayPalSDK> {
+  if (window.paypal) {
+    return window.paypal;
```
gustavs-gutmanis: This will probably not work when using multiple paypal buttons with different configs and/or client secrets.
I guess it can stay that way for now, as it seems very unlikely that someone might have that setup, but we have to keep this in mind and mention it in the docs, perhaps.

packages/scripts/src/components/front-end/payments/paypal/buttons.ts
```diff
@@ -179,156 +160,132 @@ async function initializePayPalButtons(root: Element) {
     return;
   }
 
-  const form = root.closest('form');
+  const form = root.closest('form') as HTMLFormElement | null;
```
gustavs-gutmanis: You can specify the type when calling the function. 

```ts
const form = root.closest<HTMLFormElement>('form');
```

gustavs-gutmanis (commented):
Looks way better! Good work

kjmartens (approved):

PR #2218, opened 2025-10-17
[SFT-1190]: add optgroup functionality for select and multi-select fields

gustavs-gutmanis:


kjmartens (approved):
One issue with Multiselect fields, but making a new task for that

PR #2223, opened 2025-10-21
feat(SFT-2429): added gibberish as spam blocking integration

seandelaney:
image

image

image

image

image


gustavs-gutmanis (dismissed):
Nice! 👍🏻

kjmartens (changes requested):
Doesn't appear to be working at all in my testing...

kjmartens (approved):

PR #2270, opened 2025-11-19
feat(SFT-2466): skip gibberish checks on certain fields

seandelaney:


gustavs-gutmanis (commented):
Maybe it's better to ignore this check on specific fields, such as payment fields instead?

kjmartens (approved):

PR #2271, opened 2025-11-19
fix(SFT-2490): search indexing for submissions

### LINKED ISSUE #2266: Searchindex: 100 000 000 rows of freeform submissions in the searchIndex table
### What happened?

One of our clients has a multisite website with 5 languages.
There are 72 Freeform forms with almost 200 000 submissions in the database.

These 200 000 submissions have created 100 000 000 (one hundred million) rows in the search index table.

Image

It seems that each new submission creates ~1400 rows in the searchindex table. It basically creates a row for each field that exists in freeform for all the 5 languages. So a form that has 4 fields it still creates rows in the searchindex for alot of other fields that are not used by the form.

```shell

```

1. 
2. 
3. 

### Freeform Edition

Pro

### Freeform Version

5.11.11

### Craft Version

5.8.15 

- [x] Unsure

### Previous Freeform Version

_No response_

seandelaney:
Fixes https://github.com/solspace/craft-freeform/issues/2266

Previously all fields from all forms where being added as global searchable attributes and indexed for a submission even though some fields didn't belong to the submission. Causing massive bloat in the search index. 
Now we're filtering the global search attributes, so when a submission is indexed, only its own specific form fields are included. 

First step to fixing this is to delete all rows for any submissions (adjust table names and/or prefixes Craft 5 schema):

`DELETE si FROM searchindex si INNER JOIN freeform_submissions fs ON fs.id = si.elementId;`

Second step is to update to latest Freeform version (this PR).

Third step is to reindex all submissions again:

`php craft freeform/submissions/reindex`

### DIFF
--- a/packages/plugin/src/Commands/SubmissionsController.php
+++ b/packages/plugin/src/Commands/SubmissionsController.php
@@ -102,6 +102,11 @@ class SubmissionsController extends Controller
                 'limit',
                 'siteId',
             ],
+            'reindex' => [
+                'limit',
+                'siteId',
+                'dryRun',
+            ],
         };
     }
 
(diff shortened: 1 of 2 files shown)


gustavs-gutmanis (dismissed):
Nice!

kjmartens (approved):

PR #2279, opened 2025-11-24
[SFT-2486]: FM bypass Queue (if used) for email notification testing

thejahid:


packages/plugin/src/Jobs/FreeformQueueHandler.php
```diff
@@ -64,6 +70,27 @@ public function executeAiFieldsJob(AiFieldsJobInterface $job, ?int $priority = n
...
+            return false;
+        }
+
+        // Additional verification: Request-Id should be present for legitimate Form Monitor requests
+        if (!isset($headers['X-Form-Monitor-Request-Id']) || empty($headers['X-Form-Monitor-Request-Id'])) {
+            return false;
+        }
+
+        return true;
+    }
```
gustavs-gutmanis, line 92: This should be happening in the FormMonitor bundle instead of the form queue handler. The handler should not ever know what a Form Monitor is.

I suggest adding a new event in the `queueJob()` function, that fires, and FormMonitor should listen to it, check if it's a form monitor request, and then prevent the queue from being used, and execute the job immediately instead

packages/plugin/src/Jobs/FreeformQueueHandler.php
```diff
@@ -49,9 +49,15 @@ public function queueJob(JobInterface $job, ?int $priority = null, bool $queuedE
...
+    public function executeNotificationJob(NotificationJobInterface $job, ?int $priority = null, ?bool $bypassQueue = null): void
     {
-        $this->queueJob($job, $priority, $this->settingsService->isNotificationQueueEnabled());
+        // Bypass queue for Form Monitor requests to ensure immediate email delivery
+        if (null === $bypassQueue) {
+            $bypassQueue = $this->shouldBypassQueueForFormMonitor($job);
+        }
+
+        $shouldQueue = $bypassQueue ? false : $this->settingsService->isNotificationQueueEnabled();
+        $this->queueJob($job, $priority, $shouldQueue);
```
gustavs-gutmanis, line 60: Take a look at the next comment below, this change is unnecessary.

gustavs-gutmanis (changes requested):

PR #2307, opened 2025-12-15
Fix Postgres error

brandonkelly:
This is an actual boolean column in Postgres. `true` will get automatically converted to `1` for MySQL.


gustavs-gutmanis (approved):

gustavs-gutmanis:
Thanks @brandonkelly!

PR #2321, opened 2026-01-07
fix(SFT-2554): added missing feed tables, columns and indexes on notification log table

seandelaney:
Adds back in missing tables and re-runs adding/updating columns and indexes for missing tables.


kjmartens:
> [!NOTE]
> This series of fixes applies to a single installation, so we'll conclude this and not include it in the public release of Freeform.

Issue #2350, opened 2026-01-22
TypeError when email field value is null in EmailRecipientNotifications


daanj1998:
### What happened?

# TypeError when email field value is null in EmailRecipientNotifications

## Problem

When submitting a form with an email field notification configured, if the email field value is `null` or empty, a `TypeError` is thrown because the `Recipient` constructor expects a non-null string for the `$email` parameter.

## Error Details

```
Exception 'TypeError' with message 'Solspace\Freeform\Notifications\Components\Recipients\Recipient::__construct(): Argument #1 ($email) must be of type string, null given, called in /var/www/html/vendor/solspace/craft-freeform/packages/plugin/src/Bundles/Notifications/SendListeners/EmailRecipientNotifications.php on line 69'
```

## Stack Trace

```
#0 /var/www/html/vendor/solspace/craft-freeform/packages/plugin/src/Bundles/Notifications/SendListeners/EmailRecipientNotifications.php(69): Solspace\Freeform\Notifications\Components\Recipients\Recipient->__construct()
#1 [internal function]: Solspace\Freeform\Bundles\Notifications\SendListeners\EmailRecipientNotifications->sendToRecipients()
#2 /var/www/html/vendor/yiisoft/yii2/base/Event.php(312): call_user_func()
#3 /var/www/html/vendor/solspace/craft-freeform/packages/plugin/src/Bundles/Notifications/NotificationsBundle.php(62): yii\base\Event::trigger()
#4 [internal function]: Solspace\Freeform\Bundles\Notifications\NotificationsBundle->sendNotifications()
#5 /var/www/html/vendor/yiisoft/yii2/base/Event.php(312): call_user_func()
#6 /var/www/html/vendor/solspace/craft-freeform/packages/plugin/src/Services/SubmissionsService.php(250): yii\base\Event::trigger()
#7 /var/www/html/vendor/solspace/craft-freeform/packages/plugin/src/Services/SubmissionsService.php(220): Solspace\Freeform\Services\SubmissionsService->postProcessSubmission()
#8 /var/www/html/vendor/solspace/craft-freeform/packages/plugin/src/controllers/SubmitController.php(47): Solspace\Freeform\Services\SubmissionsService->handleSubmission()
```

## Root Cause

In `EmailRecipientNotifications.php` at line 65, the code retrieves the field value:

```php
$recipient = $field->getValue();
```

[...]

PR #2351, opened 2026-01-22
Fix TypeError on null or empty fields

daanj1998:
Issue: https://github.com/solspace/craft-freeform/issues/2350


kjmartens (approved):

kjmartens:
Thank you for this @daanj1998!

We'll get it added into the next version of Freeform 🙂

PR #2358, opened 2026-01-27
fix(SFT-2576): label and value fields in user element migration

seandelaney:


gustavs-gutmanis (commented):
Not sure that the `static` prefix won't break some peoples installs who are on older php versions. 
When was it added to php?

seandelaney:
@gustavs-gutmanis I think this came from `composer run fix`

kjmartens (approved):

PR #2380, opened 2026-02-03
[SFT-2607]: Is it possible to reduce the total app size down from 5MB?

gustavs-gutmanis:


kjmartens (changes requested):
Issue noted in Jira

kjmartens (approved):

PR #2390, opened 2026-02-04
fix mailer service render string type safety

adrum:
We have some environment variables that default to empty strings. We were getting 500s because these `parseEnv` calls were return null instead of empty strings, which the `renderString` function expects a non-nullable string.


kjmartens:
Thanks for this @adrum!

We'll include it in the next release of Freeform. 🙂

kjmartens (approved):

PR #2395, opened 2026-02-05
Improve error handling during form submission

adrum:
Fix error handling in form submission by adding error logging. When an error would occur during ajax calls the backend resulted in a 400 or 500 level code, it just spins forever. This at least improve the experience by dispatching the Ajax error event and unlocking the submit button.


kjmartens:
Thanks for this @adrum. I'll have a developer check into it shortly 🙂

gustavs-gutmanis:
Thanks a lot for the PR @adrum!

Since you're not using an arrow function - `this` context is not bound correctly and would not unlock the submits, but I've implemented the change on our end and will commit and we will release it in the next release!
Thanks again for the suggestion and PR.

PR #2399, opened 2026-02-05
fix(SFT-2610): too many columns error when trying to view the CP Submissions index page

seandelaney:
Looking at the errors.txt file, the too many columns error will 100% be caused by the 34 forms and 1920 fields.

Looking at our submission query class, especially the “All submissions” call which tends to be selected by default, it is basically saying “Give me submissions across lots of forms, and also include all custom fields for every possible form.”

So for CP “All submissions” (Source=*) we dont need every field value for every form just to render the index page. Metadata only is enough to render rows. We had logic in place to skip content but it was added past where the table joins happen so had no effect.

For the CP individual submissions (Source=form:XYZ), then the query includes every field value. This is fine as we're scoping by form ID.

So just a small but big SQL query change to fix source = * and give us a big performance win.


kjmartens (changes requested):
This doesn't work correctly in my testing. It seems to kill off ALL custom field data in the Index page. Further to this, it chokes on some field types like File Uploads with an error.

kjmartens (approved):

PR #2434, opened 2026-02-20
fix(SFT-2644): migrating form handles and submission table names in Craft 5.9

seandelaney:
There was a bug in the migration script to fix tables in Craft 5.9. We were calling `Submission::generateContentTableName` class which in return was calling `StringHelper::toSnakeCase` again, so it was never actually fixing the issue.


kjmartens (approved):

kjmartens (changes requested):
Issues noted in Slack

kjmartens (approved):

PR #2444, opened 2026-03-03
fix(SFT-2610): too many columns error when trying to view the CP Orders index page with Craft Commerce installed

seandelaney:
Craft Commerce order that includes a Freeform Submissions field still produces the “Too many columns" error.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,8 @@
     "phpunit/phpunit": "^11.5.1",
     "friendsofphp/php-cs-fixer": "^3.85",
     "phpcompatibility/php-compatibility": "^9.3.5",
-    "squizlabs/php_codesniffer": "^3.13.2"
+    "squizlabs/php_codesniffer": "^3.13.2",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -34,7 +34,8 @@
     "phpunit/phpunit": "^11.5.1",
     "friendsofphp/php-cs-fixer": "^3.85",
     "phpcompatibility/php-compatibility": "^9.3.5",
-    "squizlabs/php_codesniffer": "^3.13.2"
+    "squizlabs/php_codesniffer": "^3.13.2",
+    "craftcms/commerce": "^5"
```
gustavs-gutmanis, line 38: Should we require the craft commerce to be installed?

seandelaney: Dev dependency and used for namespaces etc in our code so it doesn't freak out. We do the same in Calendar for Freeform and Blitz

gustavs-gutmanis (approved):

kjmartens (approved):

PR #2461, opened 2026-03-16
[SFT-2638]: Solspace AI

thejahid:


packages/client/src/app/pages/ai/ai.queries.ts
```diff
@@ -0,0 +1,129 @@
...
+  credits_remaining?: number | null;
+  lifetime_credits_purchased?: number | null;
+  user_email?: string | null;
+  created_at?: string | null;
+  payment_history?: {
+    package_price?: number;
+    credits?: number;
+    amount_added?: number;
+    paid_at?: string;
+  }[];
```
gustavs-gutmanis: Nested types which are used as Arrays should be separated out, and their type used.

```ts
type PaymentHistory = {
  package_price?: number;
  credits?: number;
  amount_added?: number;
  paid_at?: string;
}

type AiUsageResponse {
  // ...
  payment_history: PaymentHistory[];
}
```

packages/client/src/app/pages/ai/dashboard.tsx
```diff
@@ -0,0 +1,431 @@
...
+    return (
+      <div>
+        <Breadcrumb id="ai" label="AI" url="ai" />
+        <HeaderContainer>{translate('AI')}</HeaderContainer>
+        <DashboardWrapper>
+          <CardsGrid>
+            <Card>
+              <CardLabel>
+                <Skeleton width={80} height={10} />
+              </CardLabel>
```
gustavs-gutmanis: This is so big that it should be in its own file. 
I would also move the `forbidden` and `not found` in their own files. Actually, just one file, since they're identical, except the messaging. A single boolean prop would determine which copy to display.

packages/client/src/app/pages/ai/dashboard.tsx
```diff
@@ -0,0 +1,431 @@
...
+    if (!iso) return '—';
+    try {
+      const d = new Date(iso);
+      return Number.isNaN(d.getTime())
+        ? iso
+        : d.toLocaleDateString(undefined, { dateStyle: 'medium' });
+    } catch {
+      return iso;
+    }
+  }
```
gustavs-gutmanis: Why is this function inline here? 
It gets recreated on every re-render.
There's nothing that uses anything in the scope, it should be moved to a utility file and imported and used that way.
Also we have `date-fns` library, which should be used for formatting dates.
And please refrain from naming variables as `a`, `b`, `c`, etc.

packages/client/src/app/pages/ai/dashboard.tsx
```diff
@@ -0,0 +1,431 @@
...
+          <EmptyState>
+            <EmptyStateTitle>
+              {translate('Error loading usage')}
+            </EmptyStateTitle>
+            <p>{message}</p>
+          </EmptyState>
+        </DashboardWrapper>
+      </div>
+    );
+  }
```
gustavs-gutmanis: too many inline cases of rendering something completely different.
All of them share the same layout, so at least bring the layout wrappers out to a separate component and use that by passing only the children, that would make this file have less filler code which makes it difficult to navigate and understand.

packages/client/src/app/pages/ai/dashboard.tsx
```diff
@@ -0,0 +1,431 @@
...
+            <>
+              <Card>
+                <CardLabel>{translate('Status')}</CardLabel>
+                <CardValueSmall>
+                  {summary.mode === 'trial' && translate('Free trial')}
+                  {summary.mode === 'plan' && translate('Active plan')}
+                  {summary.mode === 'blocked' &&
+                    translate('Usage limit reached')}
+                  {(!summary.mode || summary.mode === 'unknown') &&
+                    translate('Not configured')}
```
gustavs-gutmanis: I would suggest writing a little function which does this, and then pass the `mode` to it.

```tsx
const getSummaryModeLabel = (mode: SummaryMode): string => {
  switch (mode) {
    case "trial":
      return "Free Trial";

    // ...
  }
}

// ...

<CardValueSmall>
  {getSummaryModeLabel(summary.mode)}
</CardValueSmall>
```

packages/client/src/app/pages/ai/plans.tsx
```diff
@@ -0,0 +1,398 @@
...
+  MetricsTableRow,
+  PlansHeaderActions,
+  PlansSelect,
+  Section,
+  SectionTitle,
+} from './dashboard.styles';
+
+function formatBundlePrice(
+  price: number,
+  _bundleCurrency: string,
```
gustavs-gutmanis: this should not be prefixed with `_` since it's a variable that is actually used and is not private.

packages/client/src/app/pages/ai/plans.tsx
```diff
@@ -0,0 +1,398 @@
...
+        maximumFractionDigits: 0,
+        minimumFractionDigits: 0,
+      })
+    : price.toLocaleString();
+  if (currency === 'eur') return `€${formatted}`;
+  if (currency === 'usd') return `$${formatted}`;
+  return `${formatted} ${currency.toUpperCase()}`;
+}
+
+const CURRENCY_OPTIONS: { value: string; label: string }[] = [
```
gustavs-gutmanis: Please don't inline types, it's harder to read, and easy to miss things like `[]` at the end.

packages/client/src/app/pages/ai/plans.tsx
```diff
@@ -0,0 +1,398 @@
...
+  { value: 'usd', label: 'USD' },
+  { value: 'eur', label: 'EUR' },
+];
+
+export const AiPlans: React.FC = () => {
+  const [addCreditLoading, setAddCreditLoading] = React.useState(false);
```
gustavs-gutmanis: It would be nice to import `useState` directly from `React` so that `React` is imported only as a type.


```suggestion
  import { useState } from "react";
  
  const [addCreditLoading, setAddCreditLoading] = useState(false);
```

packages/client/src/app/pages/ai/plans.tsx
```diff
@@ -0,0 +1,398 @@
...
+  const [selectedCurrency, setSelectedCurrency] = React.useState<string | null>(
+    null
+  );
+
+  const {
+    data: plans,
+    isFetching,
+    error,
+    isError,
+  } = useAiPlansQuery(selectedCurrency === null ? undefined : selectedCurrency);
```
gustavs-gutmanis: maybe we can just pass the `selectedCurrency` and handle the `null` case inside the `useAiPlansQuery` instead?

packages/client/src/app/pages/ai/plans.tsx
```diff
@@ -0,0 +1,398 @@
...
+    if (
+      bundles.length > 0 &&
+      !bundles.some((b) => String(b.key) === selectedBundleKey)
+    ) {
+      setSelectedBundleKey(defaultBundleKey);
+    }
+  }, [bundles, defaultBundleKey, selectedBundleKey]);
+
+  const isNotFound =
+    isError && axios.isAxiosError(error) && error.response?.status === 404;
```
gustavs-gutmanis: `axios.isAxiosError(error)` can be done once, instead of in every conditional.
A helper function even can be written to check for statuses on errors if they're axios errors or not, since this is used in so many places.

packages/client/src/app/pages/ai/plans.tsx
```diff
@@ -0,0 +1,398 @@
...
+  const isForbidden =
+    isError && axios.isAxiosError(error) && error.response?.status === 403;
+
+  const emptyTitle = isNotFound
+    ? translate('Solspace AI is not enabled')
+    : translate('Authorize Solspace AI to view plans');
```
gustavs-gutmanis:
```suggestion
  const emptyTitle = translate(isNotFound ? 'Solspace AI is not enabled' : 'Authorize Solspace AI to view plans');
```

packages/client/src/app/pages/ai/plans.tsx
```diff
@@ -0,0 +1,398 @@
...
+        }
+      >
+        {translate('Plans')}
+      </HeaderContainer>
+      <DashboardWrapper>
+        {plans && bundles.length > 0 ? (
+          <>
+            <Section>
+              <SectionTitle>{translate('Credit bundles')}</SectionTitle>
+              <div style={{ marginTop: '0.5rem' }}>
```
gustavs-gutmanis: Is it possible to not inline styles? Looks like these need to be a separate styled component

packages/client/src/app/pages/ai/plans.tsx
```diff
@@ -0,0 +1,398 @@
...
+                        {translate('Credits')}
+                      </MetricsTableHeaderCell>
+                    </MetricsTableRow>
+                  </MetricsTableHead>
+                  <tbody>
+                    {bundles.map((b) => (
+                      <MetricsTableRow key={b.key}>
+                        <MetricsTableCell>
+                          {formatBundlePrice(
+                            b.price,
```
gustavs-gutmanis: `b`. Please use `bundle`, or do `bundles.map(({ price, currency }` or something

packages/client/src/app/pages/forms/list/modals/modal.form.create-with-ai.tsx
```diff
@@ -0,0 +1,219 @@
...
+      label: translate('Describe your form'),
+      instructions: translate('Describe the fields and purpose of the form.'),
+      required: true,
+      rows: 4,
+      placeholder: translate(
+        'e.g. Contact form with name, email, phone, and a message box'
+      ),
+    }),
+    []
+  );
```
gustavs-gutmanis: these can be simple objects.
also - since the v5.15 changes you can pass these props directly to the `<Control>` without needing a `property` variable, so that makes it easier to use control components

packages/client/src/app/pages/forms/list/modals/modal.form.create-with-ai.tsx
```diff
@@ -0,0 +1,219 @@
...
+          'message' in res &&
+          typeof (res as { message: unknown }).message === 'string'
+        ) {
+          message = (res as { message: string }).message;
+        }
+      }
+      setError(
+        message ||
+          translate('Form generation failed. Please try again or rephrase.')
+      );
```
gustavs-gutmanis: this looks very GPT-y 😆 
it would be very nice to write a handler function for extracting error messages from AJAX errors, or just use the optimistic path and assume we know how the error looks.

packages/client/src/app/pages/forms/list/views/grid/grid.empty.tsx
```diff
@@ -73,9 +79,33 @@ export const GridEmpty: React.FC = () => {
             )}
           </p>
 
-          <button className="btn submit add icon" onClick={openCreateFormModal}>
-            {translate('Create a new Form')}
-          </button>
+          <div style={{ display: 'flex', gap: '0.5rem', flexWrap: 'wrap' }}>
```
gustavs-gutmanis: Please write a new styled component for this. Do not use inline styles

packages/plugin/src/controllers/api/AiController.php
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        $payload = array_filter([
+            'license_key' => $licenseKey,
+            'success_url' => $successUrl,
+            'cancel_url' => $cancelUrl,
+            'plan' => $body['plan'] ?? null,
+            'bundle_key' => $body['bundle_key'] ?? null,
+            'currency' => $body['currency'] ?? null,
+        ], static function ($v): bool { return $v !== null && $v !== ''; });
```
gustavs-gutmanis: This looks like a prime candidate for shorthand anonymous functions

```suggestion
        ], static fn ($v) => $v !== null && $v !== '');
```

kjmartens (approved):

PR #2496, opened 2026-04-22
fix(SFT-2740): improves the spam reason message for turnstile

seandelaney:
Integration was blindly calling Cloudflare's API even when the token is empty/null. This is not helpful and produces a confusing error like `missing-input-response` (I can replicate this part) so I have improved the messaging so users can try and better debug why a submission goes into spam folder. So now if the token is empty because `challenges.cloudflare.com` is blocked, users can tell rather than get a generic “The response parameter is missing“ message.


gustavs-gutmanis (approved):
Beautiful

kjmartens (approved):

PR #2503, opened 2026-04-28
fix(SFT-2740): improved script loading for captchas + more spam reason message info

seandelaney:
This still doesn't fix the issue we had on SFT-2740. Its more to help debug and prevent real submissions not making it to the finish line.

Servers side changes:

- If missing token then we always show an error, never silently flag submission as spam. Also a sure sign that something is blocking the token request to API.
- If the token is present but rejected by API, then respect failure behaviour setting - send to spam or display an error.
- The form handle is now included in spam reason messaging - basically if a customer sees a spam submission for contact but the reason says newsletter, there's a clear sign the captcha token from one form is being validated against another form - clean sign we have a bug somewhere else.
- I updated captcha error strings to be more descriptive. Useful when a user has multiple captchas enabled on the same form and they are fighting each other.

Client side changes:

- Captcha script loader now resolves immediately if the same script is already loaded in the DOM.Previously the promise was never resolved. We left it outside the rain.
- Removed submit button from lazy-load triggers list. This fixes a race condition I managed to replicate when testing. I was loading the form with autofill extension and clicking submit button.
- Added captcha JS polling. So whether the user fills out form fields and/or clicks submit, if the user is on a very slow connection, we now wait for the captcha scripts to load before submitting the form. Times out after 8 seconds and lets server side validation checks handle it. Basically if the JS still hasn't loaded after 8 seconds, its a sure sign its blocked by firewall etc.


packages/plugin/src/Integrations/Captchas/hCaptcha/hCaptcha.php
```diff
@@ -252,31 +263,35 @@ private function getValidationErrors(Form $form): array
...
         }
 
         if (\in_array('missing-input-response', $errorCodes, true)) {
-            $errors[] = 'The response parameter is missing.';
+            $errors[] = 'The hCaptcha response token was not submitted.';
         }
 
         if (\in_array('invalid-input-response', $errorCodes, true)) {
-            $errors[] = 'The response parameter is invalid or malformed.';
+            $errors[] = 'The hCaptcharesponse parameter is invalid or malformed.';
```
gustavs-gutmanis:
```suggestion
            $errors[] = 'The hCaptcha response parameter is invalid or malformed.';
```

packages/scripts/src/components/front-end/captchas/recaptcha/v2-checkbox.ts
```diff
@@ -35,6 +35,59 @@ const createCaptcha = (event: FreeformEvent): HTMLDivElement | null => {
...
+
+      // Wait for 8 seconds
+      if (elapsed >= 8000) {
+        clearInterval(poll);
+
+        resolve();
+      }
+    }, 100);
+  });
+};
```
gustavs-gutmanis: Maybe these could be moved to a single function that handles each case?


```suggestion
// Poll captcha until it has a token value or the timeout is reached. Resolves either way - if the token never appears, the server handles the missing token gracefully via the server validation checks.
const waitForCaptchaToken = (
  form: HTMLFormElement, 
  selector: string
): Promise<void> => {
  return new Promise<void>((resolve) => {
    let elapsed = 0;

    const poll = setInterval(() => {
      const tokenInput = form.querySelector<HTMLInputElement>(selector);
      if (tokenInput?.value) {
        clearInterval(poll);
        resolve();

        return;
      }

      elapsed += 100;

      // Wait for 8 seconds
      if (elapsed >= 8000) {
        clearInterval(poll);
        resolve();
      }
    }, 100);
  });
};
```

gustavs-gutmanis (dismissed):
Looks great!
I would probably move the `waitForToken` function into a common one used by all of the captchas, since only the selector changes.

kjmartens (approved):

PR #2523, opened 2026-05-18
refactor(SFT-2766): Performance issues when using many Conditional Email Notifications

gustavs-gutmanis:
Adding memoization, and memoizing as much as possible to reduce redundant object hydration and database calls.


kjmartens (changes requested):
See Jira issue for details

kjmartens (changes requested):
Found another issue... when loading the form builder, existing Conditional Rules (Rules tab) do not appear, and this error is in the Network Response:

`Solspace\\Freeform\\Records\\Rules\\RuleConditionRecord has no relation named \"form\".`

kjmartens (approved):

PR #2550, opened 2026-06-15
fix(SFT-2806): improve security handling and update React rendering in client app

thejahid:
- Replaced direct usage of securityKey with SecurityHelper in multiple files for better security practices.
- Updated React rendering logic in the client app to use mountFreeformApp for improved structure and maintainability.
- Removed deprecated files and updated asset paths in manifest.json to reflect new file names.


packages/client/src/app/pages/settings/settings.sidebar.tsx
```diff
@@ -21,7 +21,9 @@ export const SettingsSidebar: React.FC<Props> = ({ activeKey }) => {
   const { data, isFetching } = useQuery({
     queryKey: ["settings", "navigation"],
     queryFn: () => {
-      return axios.get("/api/settings/navigation").then((res) => res.data);
```
gustavs-gutmanis, line 24: Would `axios.get<Record<string, Item>>()` work maybe?

packages/plugin/src/Library/Helpers/EncryptionHelper.php
```diff
@@ -57,11 +57,13 @@ public static function decryptExportData(string $key, array $encryptedData): arr
 
     public static function encryptByKey(string $key, string $value): string
     {
-        return base64_encode(\Craft::$app->getSecurity()->encryptByKey($value, $key));
+        return SecurityHelper::encrypt($value, '' !== $key ? $key : null);
```
gustavs-gutmanis, line 60: I think it's best to handle empty key in the `::encrypt()` function, rather than always check before everywhere

gustavs-gutmanis (dismissed):

kjmartens (approved):

PR #2585, opened 2026-07-19
[SFT-2816]: Frontend library

thejahid:


biome.json
```diff
@@ -88,5 +90,21 @@
...
     }
-  }
+  },
+  "overrides": [
+    {
+      "includes": ["**/*.vue"],
+      "linter": {
+        "rules": {
+          "correctness": {
+            "noUnusedVariables": "off",
```
gustavs-gutmanis: Why would we leave this off?

packages/frontend/react/dist/version.js
gustavs-gutmanis, line 1: I'm not sure we should be committing in `dist/` files. We should have a CI process which builds and publishes the packages.

packages/frontend/react/src/renderers/builtin/fields.tsx
gustavs-gutmanis, line 1: It would be pretty nice to have a separate file for each field type, so it's easier to manage and oversee

gustavs-gutmanis (commented):
Overall great, with some minor questions

kjmartens (approved):

PR #2619, opened 2026-08-31
fix(SFT-3026): respect configured output date formats in Mailchimp birthday and date fields

### LINKED ISSUE #2615: Mailchimp birthday: day and month silently swapped for days 1–12
**Describe the bug**

`normalizeToBirthday()` converts some birthdays and not others.

`25/12` (25 December) throws in `new \DateTime()`, falls through to the format list, and is correctly flipped to `12/25`.

`09/06` (9 June) does not throw — PHP reads it as 6 September — so it returns immediately, unchanged. Mailchimp stores 6 September.

So on a standard MM/DD audience, day-first input works for days 13–31 and is silently reversed for days 1–12. No error, and the submission reports success, so nothing surfaces to the person filling in the form or in the logs. We had 258 birthdays stored reversed across 223 contacts before noticing.

The format list in the catch block starts with `d/m`, but for days 1–12 it's never reached.

We hit this from the other side as well: once our audience was set to DD/MM, the days 13–31 that do get flipped started coming back as `Please enter a month (01-12) and a day (01-31)`, and the whole contact was rejected rather than just the field. Same root cause.

**Steps to reproduce**

1. Map a Freeform text field to a Mailchimp birthday merge field
2. Submit `09/06`, meaning 9 June
3. The contact is stored with a birthday of 6 September

**Expected behavior**

`09/06` and `25/12` are treated consistently.

**Craft & Plugin Info**

Craft 5.8.3 / Freeform 5.11.4 Pro / upgrade

Still present in 5.15.27.

seandelaney:
Fixes https://github.com/solspace/craft-freeform/issues/2615

### DIFF
--- a/packages/plugin/src/Integrations/EmailMarketing/Mailchimp/BaseMailchimpIntegration.php
+++ b/packages/plugin/src/Integrations/EmailMarketing/Mailchimp/BaseMailchimpIntegration.php
@@ -397,12 +397,26 @@ abstract class BaseMailchimpIntegration extends EmailMarketingIntegration implem
                 continue;
             }
 
+            $options = null;
+            if (!empty($field->options)) {
+                $options = [];
+                foreach ($field->options as $key => $value) {
+                    if (\is_scalar($value)) {
+                        $options[] = [
+                            'key' => $key,
+                            'label' => (string) $value,
+                        ];
+                    }
+                }
+            }
+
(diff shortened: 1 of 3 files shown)


gustavs-gutmanis (approved):
Looks great 👌🏻

kjmartens (approved):

